PR #9, opened 2019-01-24
Update routes to full platform route

ehelms:


beav:
we may need to do something more sophisticated later, but this works fine for now.

ACK!

PR #10, opened 2019-01-25
add tests and v0 api

beav:


beav:
@johnsonm325 after this is merged, you'll need to use "/r/insights/platform/drift/v0/compare" for your URL

.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: python
 
-script:
+before_script:
```
ehelms: Is there a reason to do this as a before_script vs. just having multiple listed script entries?

ehelms: stages could also be useful: https://docs.travis-ci.com/user/build-stages/

app.py
```diff
@@ -1,44 +1,3 @@
...
-
-
-@app.route("/r/insights/platform/drift/status")
-def status():
-    return jsonify({'status': "running"})
-
-
-if __name__ == "__main__":
-    app.run(host='0.0.0.0', port=8080)
+create_app().run(port=8080)
```
ehelms: We still need the `host=0.0.0.0` don't we?

beav: oh yeah good catch

.travis.yml
beav: it should just be in `script`, I misread. `before_script` is for stuff that the computers take care of (loading deps, etc) that would show a broken build instead of a failed build that requires code fixes

beav:
@ehelms PR should be up to date at this point

drift/app.py
```diff
@@ -0,0 +1,41 @@
+from flask import Flask
+from flask_cors import CORS
+
+from drift.views import v0
+
+PHONY_DATA = {'packages': [{'name': 'gcc',
```
ehelms: Assume oyu meant to drop this?

drift/views/phony_data.py
```diff
@@ -0,0 +1,24 @@
+PHONY_DATA = {'packages': [{'name': 'gcc',
```
ehelms: A follow up would be to add in the `hosts` section of the schema that identifies each host in the comparison and it's metadata.

drift/app.py
beav: oops, ya not sure how i missed that

drift/views/phony_data.py
beav: exactly

beav: @ehelms I'll add it in a new PR, I wanted to get this one in so @johnsonm325 can change his URL

PR #11, opened 2019-01-25
add last_updated iso8601 date

beav:


beav:
@ehelms this is what I was thinking re: our other convo in the other PR. what do you think?

PR #12, opened 2019-01-28
add openapi spec

beav:


drift/openapi/api.spec.yaml
```diff
@@ -0,0 +1,36 @@
+---
+openapi: 3.0.1
+info:
+  version: "0.01"
```
ehelms: 0.01? Not seen that syntax, or is this a typo?

beav: nah, but I could put 0.1 in there or something else. Basically any number < 1 is ok:)

ehelms: But it's usually just an X.Y versioning scheme?

beav: ya you are right, I got the numbers swapped in my head https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

PR #14, opened 2019-01-28
enable debug logging

beav:


app.py
```diff
@@ -1,3 +1,7 @@
+import os
+
 from drift.app import create_app
 
-create_app().run(host='0.0.0.0', port=8080)
+drift_debug = os.getenv('DRIFT_DEBUG')
+
+create_app().run(host='0.0.0.0', port=8080, debug=drift_debug)
```
ehelms: Is this a true/false value? You could answer my question by just adding this to the README :)

beav: updated!

PR #16, opened 2019-01-29
call out to inventory svc for health check

beav:


beav:
@ehelms after this is merged, `INVENTORY_SVC_URL` will need to be set

drift/views/v0.py
```diff
@@ -1,9 +1,15 @@
...
+import os
+import requests
+from requests import RequestException
 
 from drift.views.phony_data import PHONY_DATA
 
+# TODO: manage these settings in a config.py
 APP_URL_PREFIX = "/r/insights/platform/drift"
 API_VERSION_PREFIX = "/v0"
+INVENTORY_SVC_URL = os.getenv('INVENTORY_SVC_URL', "http://inventory_svc_url_is_not_set")
```
ehelms: I suppose while the UI knows where it's rooted, the backend's may not. Looking at compliance, it looks like we can use the Openshift service hostname:

https://github.com/RedHatInsights/compliance-backend/blob/0f6d62d2fdf13ada9a5baa67b7949cdfd2ae5039/config/settings.yml#L3

Looks like this will need support to vary per environment. We can hard-code for now if it helps development, or keep it configurable and document needing to set it when you run the app.

beav: @ehelms do you have a preference between keeping it in conf files vs env var? I was going to do conf file at first but I figured there is not enough configuration yet (just one value so far) to warrant a conf file

ehelms: Environment variables work if we want to leave this generic and rely on the user/tools to configure different setups. The alternative would be to introduce an environment variable configuration for each env: ci, qe, prod. And hard code 'configs' based on those elements. I can go either way, so whatever makes your development life easier.

beav: let's stick with env var for now if thats ok

ehelms: Roger.

PR #17, opened 2019-01-31
talk to inventory service

beav:
This commit does a few things:

 * reaches out to inventory service for facts based on the IDs sent in
 * "pivots" facts from inventory service into something usable by the
    web UI
 * returns a 400 error if `host_ids[]` is not present
 * returns a 400 error `x-rh-identity` header is not present

This commit also pins Travis tests at python 3.6.


drift/views/v0.py
```diff
@@ -1,35 +1,30 @@
...
-    try:
-        r = requests.get(INVENTORY_SVC_URL + INVENTORY_SVC_PATH + '/health')
-        if r.status_code == requests.codes.ok:
-            inventory_connection_status = "pass"
-    except RequestException as re:
-        current_app.logger.warn("Unable to contact inventory service: %s" % re)
-        pass
-
-    return jsonify({'status': "running", 'inventory_connection': inventory_connection_status})
+    return jsonify({'status': "running"})
```
ehelms: Dropping the inventory status?

beav: ya, I had it there mostly as a baby step to get further on talking to the inventory service. I wasn't sure if I should leave it or not, but I figured we wouldn't want to get an alert on drift service if inventory service was down.

ehelms: Fair. We should think about if inventory service is unreachable how that affects the status of Drift given we rely heavily on inventory.

drift/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class BadRequest(Exception):
+    status_code = 400
```
ehelms: Trying to recall my python vs. ruby. Given this is a constant, should it be STATUS_CODE ?

drift/info_parser.py
```diff
@@ -0,0 +1,43 @@
+DRIFT_FACTS = {'bios_uuid', 'display_name'}
```
ehelms: Do we want to care about 'trimming' the list of facts vs. showing all available facts?

drift/info_parser.py
```diff
@@ -0,0 +1,43 @@
...
+    # TODO: we are assuming all info keys exist for each host. This is not a good assumption
+    available_info_names = ids_and_info[0].keys()
+
+    info_comparisons = [create_comparison(ids_and_info, info_name)
+                        for info_name in available_info_names
+                        if info_name in info_keys]
+    return info_comparisons
+
+
+def create_comparison(hosts_infos, info_name):
```
ehelms: The plural seems unneeded on `infos`. What is `info_name` in this respect?

drift/views/v0.py
beav: I'll add a card for this. I think as long as we return a 500 and don't retry the connection from drift to inventory it should be OK.

drift/info_parser.py
```diff
@@ -0,0 +1,43 @@
...
+        info_comparison = 'SAME'
+
+    return {'name': info_name, 'status': info_comparison, 'hosts': host_values}
+
+
+def host_metadata(host):
+    return {'id': host['id'], 'fqdn': host['fqdn'], 'last_updated': host['updated']}
+
+
+def build_comparisons(hosts_from_inventory_svc):
```
ehelms: What is svc?

drift/inventory_service_interface.py
```diff
@@ -0,0 +1,25 @@
...
+    return incoming_headers.get(AUTH_HEADER_NAME)
+
+
+def fetch_hosts(host_ids, service_auth_key):
+    auth_header = {AUTH_HEADER_NAME: service_auth_key}
+    hosts = []
+
+    for host_id in host_ids:
+        response = requests.get(INVENTORY_SVC_URL + INVENTORY_SVC_PATH +
+                                '/hosts/' + host_id, headers=auth_header)
```
ehelms: The base URL with `hosts` path could be built outside of the list iteration and potentially stored as a constant above.

drift/views/v0.py
```diff
@@ -1,35 +1,30 @@
...
 
 @section.route("/compare")
 def compare():
-    if 'X-RH-IDENTITY' in request.headers:
-        current_app.logger.debug("X-RH-IDENTITY received")
-    return jsonify(PHONY_DATA)
+    host_ids = request.args.getlist('host_ids[]')
+    auth_key = get_key_from_headers(request.headers)
+
+    if host_ids is None or host_ids == []:
```
ehelms: If I recall my python, this can be simplified to `if not host_ids`

tests/test_inventory_service_interface.py
```diff
@@ -0,0 +1,59 @@
+from drift import inventory_service_interface
+import responses
+
+import unittest
+
+
+class InventoryServiceTests(unittest.TestCase):
+
+    HOST_TEMPLATE = '''
```
ehelms: Could this be moved out to a fixture file?

drift/info_parser.py
beav: service. I guess i can flip it to be "inventory_service_hosts" or something

drift/views/v0.py
beav: (card added, marking as resolved) :card_file_box:

drift/exceptions.py
beav: This is some boilerplate from the flask docs on raising exceptions. However it still seems a little strange to me that BadRequest is something we have to define. I'll look into it more today.

drift/info_parser.py
beav: When the data comes back, it's mostly various IDs that wouldn't make sense to compare. There's a `facts` list as well but it's empty currently.

I think eventually we'll just be using the `facts` list, but for now I use this set to create some ersatz facts based on the most easily recognized IDs :blowfish:

drift/info_parser.py
ehelms: Ahh, I suck at mentally interpreting acronyms when I read code.

drift/info_parser.py
beav: it is the key for the item coming back from inventory service. I originally called it `fact_name` but not all pieces of information coming back are facts.

I'll add a docblock to these methods, they are kinda tricky to understand

drift/views/v0.py
beav: i just tried, you're right

drift/info_parser.py
beav: docblocks added

drift/inventory_service_interface.py
beav: updated

tests/test_inventory_service_interface.py
beav: done

tests/test_v0_api.py
```diff
@@ -0,0 +1,64 @@
...
+
+import mock
+import unittest
+
+
+class ApiTests(unittest.TestCase):
+
+    AUTH_HEADER = {'X-RH-IDENTITY': 'VGhlIEZhc3QgYW5kIHRoZSBGdXJpb3VzOiBUb2t5byBEcmlmdAo='}
+
+    FETCH_HOSTS_RESULT = [
```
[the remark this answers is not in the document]
beav: done

drift/info_parser.py
beav: updtd 4 fwr acrnms

tests/test_inventory_service_interface.py
ehelms: Assuming you meant to move this one as well?

drift/info_parser.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def _host_metadata(host):
+    """
+    create a metadata field for one host
+    """
+    return {'id': host['id'], 'fqdn': host['fqdn'], 'last_updated': host['updated']}
+
+
+def build_comparisons(inventory_service_hosts):
```
ehelms: Nitpick -- move this to the top since its the `public` function to make it more obvious when reading the code?

beav:
thanks @ehelms !

PR #20, opened 2019-02-06
doc and example on using `local-drift-backend.js`

beav:


README.md
```diff
@@ -1,7 +1,8 @@
 # drift-backend
-host drift analysis service
+Host drift analysis service
```
ehelms: System unless you want to fix this later?

beav: ya, will fix later as part of other card

README.md
```diff
@@ -11,3 +12,9 @@ To run:
 To set the debug level (`INFO` by default):
 
 `LOG_LEVEL=DEBUG INVENTORY_SVC_URL=<inventory service url> python3 app.py`
+
+If you would like to use this service with insights-proxy, you can use the
+included `local-drift-backend.js` like so, from the `insights-chrome/build`
+dir:
+
+`SPANDX_CONFIG=/path/to/drift-backend/local-drift-backend.js bash ../../insights-proxy/scripts/run.sh`
```
ehelms: Based on where this is running from, and the example file location, I'd assume this would just be:

```
`SPANDX_CONFIG=local-drift-backend.js bash ../insights-proxy/scripts/run.sh`
```

local-drift-backend.js
```diff
@@ -0,0 +1,3 @@
+module.exports.routes = {
+	   '/r/insights/platform/drift' : { host: 'http://localhost:8080' },
```
ehelms: Nitpick but 4 spaces indentation to match other JS.

README.md
beav: you gotta do one `..` to get out of the `build` dir, and then another to get out of `insights-chrome`. The `insights-proxy` dir is a peer to the chrome dir.

ehelms: AHh, why do you need to be in the insights-chrome/build directory?

PR #21, opened 2019-02-06
return a better error when uuid is not found

beav:
Previously, if you requested a UUID that was not found by inventory service,
you'd get a 500 error.

Instead, return a 400 error with a more helpful message.

### DIFF
--- a/drift/views/v0.py
+++ b/drift/views/v0.py
@@ -5,7 +5,7 @@ import json
 import base64
 
 from drift import info_parser
-from drift.exceptions import HTTPError
+from drift.exceptions import HTTPError, SystemNotReturned
 from drift.inventory_service_interface import fetch_hosts, get_key_from_headers
 
 APP_URL_PREFIX = "/r/insights/platform/drift"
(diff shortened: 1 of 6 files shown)


drift/views/v0.py
```diff
@@ -25,8 +25,11 @@ def compare():
     if auth_key is None:
         raise HTTPError(HTTPStatus.BAD_REQUEST, message="auth header not specified")
 
-    comparisons = info_parser.build_comparisons(fetch_hosts(host_ids, auth_key))
-    return jsonify(comparisons)
+    try:
+        comparisons = info_parser.build_comparisons(fetch_hosts(host_ids, auth_key))
+        return jsonify(comparisons)
+    except SystemNotReturned as error:
+        raise HTTPError(HTTPStatus.BAD_REQUEST, message=error.message)
```
ehelms: Does this return a string or JSON? If a string, do we want to standardize on a JSON formatted error message format? Are there any standards within the platform on error returns?

beav: it returns a json:

```
{"message":"System d1d22ce7-4e52-4767-867a-2ccac52485ec not available to display"}
```

beav: @ehelms the platform docs have some differences with the way the API is right now. I'll put in a new card for it, but it's something that is going to be bigger than this PR.

ehelms: And last point of clarity, this will 'fully error out' as in, instead of just showing any systems we can, right now it will bomb out? I think that's fine. Just making sure I follow your idea here. We should probably talk to Rox about this interaction next sprint.

beav: correct, if any UUID is not found it will bail on the whole thing. It will also error on the first UUID and not proceed with fetching the rest.

PR #23, opened 2019-02-07
rename 'compare' to 'comparison_report'

beav:
It makes more sense to "get a comparison report", thus the latter is more
resty.


ehelms:
Needs a rebase

beav:
rebase complete

PR #27, opened 2019-02-11
use openapi to define API

beav:
Previously, methods were mapped to URLs using flask annotations, and
data was validated inside the methods. This is not correct.

Instead, rely on openapi to define the method inputs. Additionally,
use the openapi definition to map URLs to methods.

This commit also handles 500 errors from the inventory service.

NOTE: the returned data from /comparison_report has changed slightly.
The old format for the list of facts was:

```
{UUID1: value1,
 UUID2: value2}
```

It is now:

```
{id: UUID1, value: value1}
{id: UUID2, value: value2}
```

The old style of having data inside of a hash key is not recommended
in the API specs I read.

### DIFF
--- a/drift/inventory_service_interface.py
+++ b/drift/inventory_service_interface.py
@@ -3,7 +3,8 @@ import os
 import requests
 from urllib.parse import urljoin
 
-from drift.exceptions import SystemNotReturned
+
+from drift.exceptions import SystemNotReturned, InventoryServiceError
 
 AUTH_HEADER_NAME = 'X-RH-IDENTITY'
 INVENTORY_SVC_HOSTNAME = os.getenv('INVENTORY_SVC_URL', "http://inventory_svc_url_is_not_set")
@@ -16,7 +17,7 @@ def get_key_from_headers(incoming_headers):
     return incoming_headers.get(AUTH_HEADER_NAME)
 
 
-def fetch_systems(system_ids, service_auth_key):
+def fetch_systems(system_ids, service_auth_key, logger):
     if len(system_ids) > MAX_UUID_COUNT:
         raise SystemNotReturned("Too many systems requested, limit is %s" % MAX_UUID_COUNT)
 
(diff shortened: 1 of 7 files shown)


drift/inventory_service_interface.py
```diff
@@ -1,9 +1,11 @@
 import json
 import os
 import requests
+from flask import current_app
```
ehelms: Is there a way to pass around the logger without requiring 'flask` ? This had a nice encapsulation before being loosely coupled.

drift/openapi/api.spec.yaml
```diff
@@ -11,21 +11,49 @@ info:
...
             application/json:
               schema:
                 $ref: "#/components/schemas/Status"
 
+  /comparison_report:
+    parameters:
+      - $ref: '#/components/parameters/rhIdentityHeader'
+    get:
+      summary: fetch comparison report
+      description: "Fetch a comparison report for given system IDs. Currently,
```
ehelms: I think we could drop `currently` here. Let's not commit to 'future' language if we can avoid it.

drift/inventory_service_interface.py
beav: ya, I could send it into `fetch_systems`. The other option is to not log what the result is and assume we can just look at the logs on the other service and correlate with timestamps.

ehelms:
Is there an example of the output in the docs from comparison_report or in the tests?

beav:
> Is there an example of the output in the docs from comparison_report or in the tests?

If the app is deployed, you can go to `drift/v0/ui/` which has an example. However that doesn't help much when reviewing PRs or you don't have access to the /ui console  :cactus: 

I checked in an example at https://raw.githubusercontent.com/RedHatInsights/drift-backend/master/drift/openapi/comparison-report-example.json as well, and the openapi docs point to that.

drift/openapi/api.spec.yaml
```diff
@@ -34,3 +66,70 @@ components:
...
+    ComparisonReport:
+      required:
+        - facts
+        - header_mappings
+      properties:
+        facts:
+          type: array
+          items:
+            $ref: "#/components/schemas/FactComparison"
+        header_mappings:
```
ehelms: What is the idea behind `header_mappings` as the naming for this field? I don't find it obvious what it means as a data point compared to something like `systems`.

beav: I chose the name only because the data is likely to be used for the column headers in the frontend. I can change it to "systems" or something, that may make more sense

beav:
the example json will also need updating. I'll either make a new PR after this is merged, or figure out a way later to autogenerate the example

drift/openapi/api.spec.yaml
ehelms: Yea, I like to avoid if possible, API data that is "ui" specific in it's naming. When it's a public API, I think its better if it reflects what things actually are.

ehelms:
UI change -- https://github.com/RedHatInsights/drift-frontend/pull/36

ehelms:
Merging this and the UI piece

PR #29, opened 2019-02-12
Use system facts from 'facts' dict

beav:
Previously, we massaged system data to pull a few facts based on
identification data like bios UUID and fqdn.

Instead, use the actual facts dictionary coming back from inventory
service.

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -1,34 +1,53 @@
-DRIFT_FACTS = {'bios_uuid', 'display_name'}
-DRIFT_SYSTEM_ID = {'id'}
+SYSTEM_ID = 'id'
+FACT_NAMESPACE = 'inventory'
 
 
 def build_comparisons(inventory_service_systems):
     """
     given a list of system dicts, return a dict of comparisons, along with a dict of metadata.
     """
-    fact_comparison = _select_applicable_info(DRIFT_FACTS, inventory_service_systems)
+    fact_comparison = _select_applicable_info(inventory_service_systems)
 
     metadata = [_system_metadata(system) for system in inventory_service_systems]
     return {'facts': fact_comparison, 'metadata': metadata}
 
 
-def _select_applicable_info(info_keys, systems):
+def _select_applicable_info(systems):
     """
     Take a list of systems, and output a "pivoted" list of facts, where each fact
     key has a dict of systems and their values. This is useful when comparing
     facts across systems.
     """
-    # create dicts of system + info
-    ids_and_info = [{key: system[key] for key in (info_keys | DRIFT_SYSTEM_ID & system.keys())}
-                    for system in systems]
+    # create dicts of id + info
+    ids_and_info = [_system_facts_and_id(system) for system in systems]
 
     # TODO: we are assuming all info keys exist for each system. This is not a good assumption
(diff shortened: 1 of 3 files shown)


drift/info_parser.py
```diff
@@ -1,36 +1,43 @@
...
-def _select_applicable_info(info_keys, systems):
+def _select_applicable_info(systems):
     """
     Take a list of systems, and output a "pivoted" list of facts, where each fact
     key has a dict of systems and their values. This is useful when comparing
     facts across systems.
     """
     # create dicts of system + info
-    ids_and_info = [{key: system[key] for key in (info_keys | DRIFT_SYSTEM_ID & system.keys())}
+    ids_and_info = [{**{key: _system_facts(system)[key] for key in _system_facts(system)},
```
ehelms: What does `**` do in this context? This is quite the list comprehension to mentally unpack and understand.

drift/info_parser.py
```diff
@@ -1,36 +1,43 @@
...
-                        if info_name in info_keys]
+                        if info_name is not SYSTEM_ID]
     return info_comparisons
 
 
+def _system_facts(system):
+    """
+    Pull the system facts dict out from a system record
+    """
+    return system['facts'][0]['facts']
```
ehelms: Why only the first element of the array?

drift/info_parser.py
beav: It "unzips" the dicts so they are not nested when a new one is created. For example:
```
In [1]: x = {'foo': 'bar'}

In [2]: y = {'baz': 'quux'}

In [3]: {**x, **y}
Out[3]: {'foo': 'bar', 'baz': 'quux'}
```

I could also do `{key: _system_facts(system)[key]}[SYSTEM_ID] = system['id']`

Basically I just need the system ID added in alongside the rest of the facts for the transformation later on. I think I can rewrite this cleaner, gimme a few

drift/info_parser.py
```diff
@@ -1,35 +1,52 @@
...
+                        if info_name is not SYSTEM_ID]
     return info_comparisons
 
+def _find_facts_for_namespace(system, namespace):
+    """
+    return the facts for the given namespace
+    """
+    for facts in system['facts']:
+        if facts['namespace'] == namespace:
+            return facts['facts']
```
beav: I was tempted to use `reduce` here but the for loop is considered more pythonic

beav:
@ehelms updated to avoid wild list comprehension and to avoid hardcoding of first value in facts array

drift/info_parser.py
```diff
@@ -1,35 +1,52 @@
-DRIFT_FACTS = {'bios_uuid', 'display_name'}
-DRIFT_SYSTEM_ID = {'id'}
+SYSTEM_ID = 'id'
+FACT_NAMESPACE = 'inventory'
```
ehelms: We are likely to not want to restrict ourselves to a single (or any) namespace given the set of facts we'll want. If this affects your code design. If you feel this is fine for now and easy to come back and adjust as we know more then :+1:

ehelms: Just saw https://github.com/RedHatInsights/drift-backend/pull/29/files#diff-8b7de392a44280d3d4ceb46a196e26f9R44 which answers my question.

PR #31, opened 2019-02-15
Handle facts that are only on some systems

beav:
Previously, we assumed that all systems have the same set of facts. However,
it's very likely that systems have different sets of facts defined.

This commit adds additional mock facts, and adds a new fact that only appears
if the system UUID starts with a lowercase letter.

Additionally, the comparison now allows for facts that are not defined on all
systems. If a fact is only defined on some systems, the comparison will say
"INCOMPLETE_DATA" as the value for systems where the fact is missing.

Note that if a system really does have a system fact with the value of
"INCOMPLETE_DATA", the comparison report may show the incorrect state. I don't
think there is a way around this, since any value we choose needs to be
serialized to a string for the JSON object.

This commit also creates a `constants.py` which contains all constants. There
were a couple of strings that I did not move into constants (for example,
`"facts"`). It was a bit of a judgment call on which should move and which
shouldn't.

### DIFF
--- a/drift/constants.py
+++ b/drift/constants.py
@@ -1,0 +1,13 @@
+API_VERSION_PREFIX = "/v0"
+APP_URL_PREFIX = "/r/insights/platform/drift"
+AUTH_HEADER_NAME = 'X-RH-IDENTITY'
+FACT_NAMESPACE = "inventory"
+INVENTORY_SVC_HOSTS_ENDPOINT = '/r/insights/platform/inventory/api/v1/hosts/%s?per_page=%s'
+MAX_UUID_COUNT = 20
+MOCK_FACT_NAMESPACE = 'mockfacts'
+MOCK_FACTS_FILE = 'drift/mock_data/mockfacts.json'
+SYSTEM_ID_KEY = 'id'
+
+COMPARISON_SAME = "SAME"
+COMPARISON_DIFFERENT = "DIFFERENT"
+COMPARISON_INCOMPLETE_DATA = "INCOMPLETE_DATA"
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -1,4 +1,5 @@
-SYSTEM_ID_KEY = 'id'
+from drift.constants import SYSTEM_ID_KEY, COMPARISON_SAME
+from drift.constants import COMPARISON_DIFFERENT, COMPARISON_INCOMPLETE_DATA
 
 
 def build_comparisons(inventory_service_systems, fact_namespace):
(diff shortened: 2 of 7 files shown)


beav:
When I tested, the frontend handled this correctly without alteration

drift/constants.py
```diff
@@ -0,0 +1,14 @@
+API_VERSION_PREFIX = "/v0"
+APP_URL_PREFIX = "/r/insights/platform/drift"
+AUTH_HEADER_NAME = 'X-RH-IDENTITY'
+FACT_NAMESPACE = "inventory"
+INVENTORY_SVC_HOSTS_ENDPOINT = '/r/insights/platform/inventory/api/v1/hosts/%s?per_page=%s'
+MAX_UUID_COUNT = 20
+MOCK_FACT_NAMESPACE = "mockfacts"
+MOCK_FACT_NAMESPACE = 'mockfacts'
```
ehelms: This is here twice.

drift/constants.py
```diff
@@ -0,0 +1,14 @@
+API_VERSION_PREFIX = "/v0"
+APP_URL_PREFIX = "/r/insights/platform/drift"
+AUTH_HEADER_NAME = 'X-RH-IDENTITY'
+FACT_NAMESPACE = "inventory"
```
ehelms: Do we have a card for handling multiple namespaces? I can forsee that being needed.

beav: added

drift/info_parser.py
```diff
@@ -21,8 +22,10 @@ def _select_applicable_info(systems, fact_namespace):
     # create dicts of id + info
     ids_and_info = [_system_facts_and_id(system, fact_namespace) for system in systems]
 
-    # TODO: we are assuming all info keys exist for each system. This is not a good assumption
-    available_info_names = ids_and_info[0].keys()
+    # union the keys into one big set
+    available_info_names = set()
+    for fact_ids in ids_and_info:
+        available_info_names |= fact_ids.keys()
```
beav: I tried a set comprehension for this, but `keys()` is an iterator and needs to be unioned to convert it to a set. The for loop was more readable.

drift/constants.py
```diff
@@ -0,0 +1,13 @@
...
+INVENTORY_SVC_HOSTS_ENDPOINT = '/r/insights/platform/inventory/api/v1/hosts/%s?per_page=%s'
```
beav: this will eventually get broken out into config via env var, per api docs

PR #32, opened 2019-02-18
Enable prometheus and gunicorn

beav:
Previously, the drift backend app ran with flask's default single-worker
server. This works fine for development, but is not suitable for production.
For example, at startup, the app prints:

```
WARNING: Do not use the development server in a production environment.
Use a production WSGI server instead.
```

This commit enables gunicorn, which is a small webserver that can serve WSGI
apps. The built-in flask server is no longer used.

The best way to run the app now is to invoke `./run_app.sh`. Unit tests can be
run with `run_unit_tests.sh`.

Additionally, this commit enables Prometheus. We now keep stats on the following:

 * number of exceptions raised when calling `/comparison_report`
 * response time on `/comparison_report` as histogram
 * response time when calling inventory service as histogram

The stats are cleared when the app is restarted.

The health related APIs have been split out into `/mgmt/v0/`. It didn't seem to
make sense to version the health check API in the same way as the application
API. This may or may not make it easier to hide the status check as well, I am
not sure. The new API is defined under `drift/openapi/mgmt_api.spec.yaml`, and
has its own Swagger UI.


beav:
@ehelms this PR should make zero change to the app's behavior. Nothing related to `/comparison_report` was touched

wsgi.py
```diff
@@ -0,0 +1,6 @@
+from drift.app import create_app
```
ehelms: I tend to be confused by `wsgi.py` vs. `app.py` since wsgi just represents the interface vs. `app.py` being a clear entry point to me no matter your knowledge.

run_app.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if [ -z "$LOG_LEVEL" ];
+  then LOG_LEVEL='info'
+fi
+
+if [ -z "$NUM_WORKERS" ];
+  then NUM_WORKERS=2
+fi
+
+prometheus_multiproc_dir=$TEMPDIR gunicorn wsgi -w $NUM_WORKERS -b 0.0.0.0:8080 --log-level=$LOG_LEVEL --access-logfile=-
```
ehelms: What owns or controls the `prometheus_multiproc_dir` environment variable?

wsgi.py
beav: updated to make filenames clearer. `app.py` is "the app" now, and `standalone_flask_server.py` is the nonstandard way

run_app.sh
beav: It works differently than what we discussed on IRC. `prometheus_multiproc_dir` is controlled by the bash script, which is also in charge of making sure the temp dir exists and is cleaned up afterwards.

If we wanted to have the same stats across pods, we would need a shared volume for this dir. However I think for this PR it's OK to assume single-pod.

ehelms: Controlled by the bash script I get, but *what* code uses it? One of the libraries? And has to be set before the app initializes?

beav: ah I see what you mean. The metrics display code has:

```python
def metrics():
    registry = CollectorRegistry()
    multiprocess.MultiProcessCollector(registry)
    prometheus_data = generate_latest(registry)
    return prometheus_data
```

The `multiprocess.MultiProcessCollector(registry)` line initializes a collector that requires the env var. If it's not set, then the app will raise an error at startup.

ehelms: Is it possible to set that either here https://github.com/RedHatInsights/drift-backend/pull/32/files#diff-3f41e546893dc64b71aaacad12cad815R3 or before the `application.run()`

beav: I agree its kind of a strange place to initialize a registry. It's in the example but that might be just to make the example more simple. I'll see what it's doing exactly and if it can be moved so it's not initialized over and over

beav: I thought I was going to have to write a singleton (ugh), but it looks like I can throw the registry on the app context maybe. Trying that now.

beav: @ehelms registry has been put onto the app context so it's not initalized over and over

beav:
```
$ ./run_unit_tests.sh
./run_unit_tests.sh: line 8: nosetests-3: command not found
The command "./run_unit_tests.sh" exited with 0.
```

whoops i guess i better fix that

beav:
> ```
> $ ./run_unit_tests.sh
> ./run_unit_tests.sh: line 8: nosetests-3: command not found
> The command "./run_unit_tests.sh" exited with 0.
> ```
> 
> whoops i guess i better fix that

```
2.89s$ ./run_unit_tests.sh
.................
Name                                   Stmts   Miss  Cover
----------------------------------------------------------
drift/__init__.py                          0      0   100%
drift/app.py                              18      0   100%
drift/config.py                            4      0   100%
drift/constants.py                        12      0   100%
drift/error.py                             6      0   100%
drift/exceptions.py                       13      0   100%
drift/info_parser.py                      32      0   100%
drift/inventory_service_interface.py      29      0   100%
drift/metrics.py                           4      0   100%
drift/metrics_registry.py                  7      0   100%
drift/mgmt_views/__init__.py               0      0   100%
drift/mgmt_views/v0.py                     9      0   100%
drift/mock_data/__init__.py                0      0   100%
drift/mock_data/mock_data.py              25      1    96%
drift/views/__init__.py                    0      0   100%
drift/views/v0.py                         36      0   100%
----------------------------------------------------------
TOTAL                                    195      1    99%
----------------------------------------------------------------------
Ran 17 tests in 2.384s
OK
The command "./run_unit_tests.sh" exited with 0.
```

@ehelms fixed

ehelms:
This will require us to change the deployment config to run this new command given we are straying from the s2i expectations. I think you could still comply with that in the end, take a look at:

https://github.com/sclorg/s2i-python-container/blob/master/3.5/s2i/bin/run#L77

ehelms:
So you can add the APP_CONFIG value with this PR:

```
        - env:
            - name: APP_CONFIG
              value: path_to_file
```

In this section https://github.com/RedHatInsights/drift-backend/blob/master/openshift/templates/drift-backend.dc.yaml#L30

beav:
> So you can add the APP_CONFIG value with this PR:
> 
> ```
>         - env:
>             - name: APP_CONFIG
>               value: path_to_file
> ```
> 
> In this section https://github.com/RedHatInsights/drift-backend/blob/master/openshift/templates/drift-backend.dc.yaml#L30

Does the change look correct? I wasn't sure if the new value was supposed to be a peer to `image` or under `image`.

openshift/templates/drift-backend.dc.yaml
```diff
@@ -27,6 +27,9 @@ spec:
     spec:
       restartPolicy: Always
       containers:
+        - env:
```
ehelms: this is a peer to `image` so you can drop the `-` on `image` below

ehelms:
Needs a rebase

PR #35, opened 2019-02-20
disable CORS

beav:
Previously, we had CORS enabled as part of the week 1 sprint. I don't remember
exactly, but I think we did this before we had insights-proxy and the CI
environment wired up.

This appears to not be needed anymore since all requests go through the same
hostname, so its OK to disable again.


beav:
:beer:

PR #36, opened 2019-02-21
Add prometheus emptyDir volume

ehelms:


beav:
can you add the name/value for the env var to point to this as well?

name is `prometheus_multiproc_dir` (no caps), value is the mount location

ehelms:
Slightly confused, why do I want to make this an environment variable? That allows someone or something to come in and change the variable outside of the deployment configuration and break the application. The only real way to dynamically set the volumeMount path is for me to turn it into an Ansible variable and then it can be changed on deployment.

beav:
> Slightly confused, why do I want to make this an environment variable? That allows someone or something to come in and change the variable outside of the deployment configuration and break the application. The only real way to dynamically set the volumeMount path is for me to turn it into an Ansible variable and then it can be changed on deployment.

The app is only aware of the path location via the env var, it's not something I can set easily since its cooked into the library. I was thinking the path would just be listed twice in this yaml file, once for the env var (so prometheus_client knows where to write to) and once in the volume definition

ehelms:
Updated

PR #39, opened 2019-02-22
Use pipenv instead of requirements.txt

beav:
Previously, we kept track of python deps via requirements.txt and
test-requirements.txt. This worked, but did not perform version locking. It
also requires the user to remember to run virtualenv (a step I always forget).

Instead, use pipenv. This creates a Pipfile and Pipfile.lock, so we won't get
an inadvertent library upgrade during a deployment. Additionally, virtualenvs
are handled more cleanly.

This commit migrates to pipenv and also sets a new env var so s2i knows that
pipenv is in use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,5 @@ script:
   - ./run_unit_tests.sh
 
 install:
-  - pip install -r requirements.txt
-  - pip install -r test-requirements.txt
+  - pip install pipenv
+  - pipenv install --dev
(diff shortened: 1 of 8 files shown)


beav:
:tractor: confirmed that unit tests all ran w/ travis and pipenv :tractor:

beav:
@ehelms I tested this with a debug pod and it worked ok, but I did not do a full s2i build. However I ran some of the commands that s2i does and I think it will work out of the box.

PR #41, opened 2019-02-25
README updates

beav:
The readme file was missing a tag and did not render. I also added instructions
on creating a temp dir.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,21 @@ To set up pipenv:
 ```
 yum install -y pipenv
 pipenv install # will pull in deps and create virtualenv, and will print next steps to run
-``
+```
 
-To run:
+to set up temp directory (command will return a temp dir that you'll use later)
+```
+mktemp -d
+```
 
-`prometheus_multiproc_dir=/tmp/tempdir INVENTORY_SVC_URL=<inventory service url> run_app.sh`
+To run (inventory service URL can be obtained from drift or platform team):
+
+`prometheus_multiproc_dir=`mktemp -d` INVENTORY_SVC_URL=<inventory service url> ./run_app.sh`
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,16 +9,21 @@ To set up pipenv:
...
+to set up temp directory (command will return a temp dir that you'll use later)
+```
+mktemp -d
+```
 
-To run:
+To run (inventory service URL can be obtained from drift or platform team):
 
-`prometheus_multiproc_dir=/tmp/tempdir INVENTORY_SVC_URL=<inventory service url> run_app.sh`
+`prometheus_multiproc_dir=<temp dir> INVENTORY_SVC_URL=<inventory service url> ./run_app.sh`
```
ehelms: I think you could simplify this to:

```
prometheus_multiproc_dir=`mktemp -d` INVENTORY_SVC_URL=<inventory service url> ./run_app.sh
```

johnsonm325: But then, aren't you just creating a new temp directory every time? I would see that and think I needed to run it with mktemp -d every time if I didn't know any better. I guess everyone using our app would know better.

This is just my 2 cents.

beav: > But then, aren't you just creating a new temp directory every time? I would see that and think I needed to run it with mktemp -d every time if I didn't know any better. I guess everyone using our app would know better.
> 
> This is just my 2 cents.

I could go either way on it. It should be OK to create a new one every time it's run, since it's not being run over and over again programatically. The tmpwatch cleaner should clean up anything that hangs around for awhile.

ehelms: Plus you gotta remember the tempdir you made everytime. I like simple, one command run commands for ease of use.

beav:
@ehelms @johnsonm325 updated, will squash when merging

PR #42, opened 2019-02-25
Set build config resource limit to 512Mi

ehelms:


beav:
512Mi ought to be enough for anyone

PR #43, opened 2019-03-04
use pandas to flatten fact dict

beav:
The inventory service allows for fact values that are dictionaries.
This is difficult to display in the drift web UI.

This commit uses the pandas library which allows for easy dict
flattening.


beav:
:panda_face:

beav:
it seemed a little heavyweight at first to pull in pandas for this, but I'd rather use a well-tested lib than write my own dictionary flattener. Also I think pandas may be useful going forward for other fact slicing/dicing.

PR #44, opened 2019-03-04
Add liveness probe

ehelms:


openshift/templates/drift-backend.dc.yaml
```diff
@@ -31,13 +31,23 @@ spec:
...
             - name: APP_NAME
-              value: drift
+              value: "{{ app_name }}"
             - name: ENABLE_PIPENV
               value: true
             - name: prometheus_multiproc_dir
               value: /prometheus-data
+          readinessProbe:
+            httpGet:
+              path: "{{ path_prefix }}/{{ app_name }}/status"
```
beav: This is just `/mgmt/v0/status`, I moved the status and metrics APIs to a different namespace that's not under the app's main namespace.

ehelms: Can you clarify a bit? I am just changing the `status` part to `/mgmt/v0/status` such that it would be in total:

/drift/mgmt/v0/status

beav: ya, for example, on my system this works:

`http://localhost:8080/mgmt/v0/status`

ehelms: How does that differentiate itself from other applications? Doesn't this need to be namespaced?

beav: It works similar to inventory svc, https://github.com/RedHatInsights/insights-host-inventory#deployment

ehelms:
Updated with the suggested URL

beav:
@ehelms approved. I don't know if this needs anything special done in openshift after merging, will leave hitting the green button to you

PR #48, opened 2019-03-12
flatten kernel module list

beav:
The kernel module list was previously presented as a large string. This commit
breaks it apart into smaller facts that can be compared by the frontend app.

This commit also sorts all of the facts by key, which makes the web UI easier
to read.

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -12,7 +12,9 @@ def build_comparisons(inventory_service_systems, fact_namespace):
     fact_comparison = _select_applicable_info(inventory_service_systems, fact_namespace)
 
     system_mappings = [_system_mapping(system) for system in inventory_service_systems]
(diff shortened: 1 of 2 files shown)


ehelms:
Should this just flatten all Lists for now vs. flattening specific lists? The former seems more future proof and easy to remove if we ever wanted to go back to providing lists.

beav:
> Should this just flatten all Lists for now vs. flattening specific lists? The former seems more future proof and easy to remove if we ever wanted to go back to providing lists.

I did it this way at first, but each list is handled a little differently. For example, a kernel module is "loaded", but an lscpu flag is "enabled". I then wrote a flatten method where you could pass in a verb and realized that we don't want to flatten the hostname list in the same way.

Each list ended up having its own small rules, so I ended up just doing it like this. If we add more facts and a pattern emerges on how to flatten, I'm happy to genericize it.

PR #52, opened 2019-03-14
Require Python 3.6

beav:
Previously, we required Python 3.7. Some systems do not have 3.7
available, and we do not use any 3.7-specific features.

This commit bumps the requirement down to Python 3.6.


beav:
:snake:

PR #56, opened 2019-03-15
Support package profiles

beav:
This commit reads the 'installed_packages' fact and splits it out into
one fact per package. The package name is in the fact name, and the VRA
is the value. For right now, the epoch is unused.

Note that this will change when inventory service is updated to support
installed packages.


ehelms:
Needs a rebase

beav:
rebased

PR #57, opened 2019-03-18
handle rename of `cpu_flags` to `cpu.cpu_flags`

beav:
There is an open PUP PR to rename this field, since we added a few
additional CPU facts.

Additionally, we convert everything to strings. This prevents truthy
fact values like int(1) from turning into `True`.


ehelms:
Needs a rebase

beav:
rebased

PR #58, opened 2019-03-19
use new API path

beav:


ehelms:
This probably needs a UI PR to go with it.

beav:
@ehelms we are working on that part now, there are a couple of pieces that need updating.

PR #62, opened 2019-03-25
Create spandx conf for back+front

beav:
Previously, the spandx config only covered the backend app.

This commit creates a new spandx conf for backend+frontend, as well as
a backend-only conf.


local-drift.js
```diff
@@ -0,0 +1,15 @@
...
+// see https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
+const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docker.internal';
+
+module.exports = {
+    routes: {
+        '/api/drift' : { host: 'http://localhost:8080' },
+        '/apps/drift': { host: `http://${localhost}:8002` },
+        '/beta/drift': { host: `http://${localhost}:8002` },
+        '/insights/drift': { host: `http://${localhost}:8002` },
+        '/apps/chrome': { host: 'https://ci.cloud.paas.upshift.redhat.com' }
```
ehelms: Internal URL, should we be committing that?

beav:
@ehelms it looks like the spandx.js isn't used anymore in the starter app, https://github.com/RedHatInsights/insights-frontend-starter-app/commit/5fd260ebf690097b50fa745eb35989591d330c79.

I am closing this PR and will put in a card to update our stuff to use the new starter app. It appears that the new way will route to the API on its own.

PR #64, opened 2019-03-28
fix unit tests

beav:
The unit test run script ended with `rm`, which almost always returns
zero. This meant that travis thought tests were passing when they
failed.

This commit fixed the test script, and also fixes the broken test.


beav:
test should fail for this, updated script but not test to confirm failure

PR #67, opened 2019-04-02
Use new `system_profile` field

beav:
This commit uses the new `system_profile` field. It also introduces
grouping via `.`.

### DIFF
--- a/drift/constants.py
+++ b/drift/constants.py
@@ -1,7 +1,8 @@
 API_VERSION_PREFIX = "/v0"
 AUTH_HEADER_NAME = 'X-RH-IDENTITY'
 FACT_NAMESPACE = "system_profile"
-INVENTORY_SVC_HOSTS_ENDPOINT = '/api/inventory/v1/hosts/%s?per_page=%s'
+INVENTORY_SVC_SYSTEMS_ENDPOINT = '/api/inventory/v1/hosts/%s?per_page=%s'
+INVENTORY_SVC_SYSTEM_PROFILES_ENDPOINT = '/api/inventory/v1/hosts/%s/system_profile?per_page=%s'
 MAX_UUID_COUNT = 20
 MOCK_FACT_NAMESPACE = 'mockfacts'
 MOCK_FACTS_FILE = 'drift/mock_data/mockfacts.json'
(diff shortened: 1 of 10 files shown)


beav:
tests should pass now

beav:
I get 96% with local coverage, not sure why its different

```
Name                                   Stmts   Miss  Cover   Missing
--------------------------------------------------------------------
drift/__init__.py                          0      0   100%
drift/app.py                              20      0   100%
drift/config.py                            7      0   100%
drift/constants.py                        16      0   100%
drift/error.py                             6      0   100%
drift/exceptions.py                       17      2    88%   44-45
drift/info_parser.py                      62      6    90%   63, 70, 79-80, 95, 99
drift/inventory_service_interface.py      43      0   100%
drift/metrics.py                           4      0   100%
drift/metrics_registry.py                 16      1    94%   25
drift/mgmt_views/__init__.py               0      0   100%
drift/mgmt_views/v0.py                     9      0   100%
drift/mock_data/__init__.py                0      0   100%
drift/views/__init__.py                    0      0   100%
drift/views/v0.py                         34      0   100%
--------------------------------------------------------------------
TOTAL                                    234      9    96%
----------------------------------------------------------------------
Ran 16 tests in 1.903s
```

ehelms:
Codecov is a bit obtuse with its reporting.

drift/constants.py
```diff
@@ -10,3 +11,11 @@
...
 COMPARISON_DIFFERENT = "DIFFERENT"
 COMPARISON_INCOMPLETE_DATA = "INCOMPLETE_DATA"
+
+SYSTEM_PROFILE_INTEGERS = {'number_of_cpus', 'number_of_sockets',
+                           'cores_per_socket', 'system_memory_bytes'}
+SYSTEM_PROFILE_STRINGS = {'infrastructure_type', 'infrastructure_vendor',
+                          'bios_vendor', 'bios_version', 'bios_release_date',
+                          'os_release', 'os_kernel_version', 'arch', 'last_boot_time'}
+SYSTEM_PROFILE_LISTS_OF_STRINGS_ENABLED = {'cpu_flags', 'kernel_modules', 'enabled_services'}
+SYSTEM_PROFILE_LISTS_OF_STRINGS_INSTALLED = {'installed_services'}
```
ehelms: Why do we need to call out every field vs. just processing all fields presented by the system profile API?

beav: I split them out based on how they are parsed. For example, I have to cast `str()` on integers, and I have to do some additional parsing on `installed_services` so they show up as `service: installed` and not `service: true`

I could probably get rid of these sets for one-offs like `installed_services`. I didn't use a constant for `installed_packages` either.

ehelms: Thanks for the explanation! This sounds good

PR #68, opened 2019-04-03
enable grouping

beav:
This looks at the top-level facts, and if any have `.`, creates a group.


beav:
@johnsonm325 this should contain the changes we discussed earlier today. Try it out and let me know what you think

PR #70, opened 2019-04-08
rename v0 to v1 api

beav:
The API is stablized enough to be v1.  The v0 API is no longer supported.


beav:
@johnsonm325 this will require a frontend change to support

beav:
i'm about to rename a field, do not merge this yet.

PR #71, opened 2019-04-09
UUID validate

beav:


beav:
I'll rebase once we cut over to v1 api

drift/views/v1.py
```diff
@@ -3,15 +3,28 @@
...
 
 from drift import info_parser, metrics
 from drift.exceptions import HTTPError, SystemNotReturned
 from drift.inventory_service_interface import fetch_systems_with_profiles, get_key_from_headers
 
-
 section = Blueprint('v1', __name__)
 
 
+def _uuid_test(system_ids):
```
ehelms: I sometimes see this type of function written with `validate` or `assert` language rather than `test` (which I think can get confusing with actual tests).

PR #81, opened 2019-04-25
check x-rh-header for entitlement status

beav:


drift/views/v1.py
```diff
@@ -46,6 +46,16 @@ def comparison_report():
         raise HTTPError(HTTPStatus.BAD_REQUEST, message=error.message)
 
 
+def _is_mgmt_url(path):
+    """
+    small helper to test if URL is for management API.
+    """
+    if path.startswith('/mgmt/'):
```
ehelms: `startswith` returns a boolean so you can simplify this to `return path.startswith('/mgmt')` I think.

drift/views/v1.py
```diff
@@ -57,6 +67,30 @@ def ensure_account_number():
...
+def ensure_entitled():
+    """
+    check if the request is entitled. We run this on all requests and bail out
+    if the URL is whitelisted.
+    """
+    # TODO: Blueprint.before_request was not working as expected, using
+    # before_app_request and checking URL here instead.
+    if _is_mgmt_url(request.path):
+        return  # allow request
+    auth_key = get_key_from_headers(request.headers)
```
ehelms: An empty line above here would help me with readability.

drift/views/v1.py
```diff
@@ -57,6 +67,30 @@ def ensure_account_number():
...
+@section.before_app_request
+def ensure_entitled():
+    """
+    check if the request is entitled. We run this on all requests and bail out
+    if the URL is whitelisted.
+    """
+    # TODO: Blueprint.before_request was not working as expected, using
+    # before_app_request and checking URL here instead.
+    if _is_mgmt_url(request.path):
+        return  # allow request
```
ehelms: Does python's `return` return `true` by default?

drift/views/v1.py
```diff
@@ -57,6 +67,30 @@ def ensure_account_number():
...
+    auth_key = get_key_from_headers(request.headers)
+    if auth_key:
+        entitlements = json.loads(base64.b64decode(auth_key)).get('entitlements', {})
+        if 'smart_management' in entitlements:
+            if entitlements['smart_management'].get('is_entitled'):
+                current_app.logger.debug("enabled smart management entitlement found on header")
+                return  # allow request
+    else:
+        current_app.logger.debug("identity header not sent for request")
+    current_app.logger.debug("smart management entitlement not found for account.")
```
ehelms: A line above here would help, I thought this was just mis-indented.

ehelms: Or maybe this should be in the else?

drift/views/v1.py
beav: it returns `None`. If you return a value, Flask will return that to the user instead of doing further processing.

http://flask.pocoo.org/docs/1.0/api/#flask.Flask.before_request has more info

PR #90, opened 2019-05-16
Add autoscaler config

ehelms:


openshift/roles/drift-backend/tasks/main.yml
```diff
@@ -4,6 +4,11 @@
     definition: "{{ lookup('template', 'templates/deployment-config.yaml') | from_yaml }}"
     host: "{{ openshift_host }}"
 
+- name: horizontal pod autoscaler
+  k8s:
+    definition: "{{ lookup('template', 'templates/hpa.yaml') | from_yaml }}"
```
beav: I don't see an `hpa.yaml`. Should it be in this commit?

ehelms: Ahh yep, forgot to git add. Now its there.

PR #91, opened 2019-05-17
call inventory service differently

beav:
Previously, we would send inventory service *all* of our UUIDs, and
ask for a paginated result. This resulted in two issues:

 * we were limited by the length of the URL
 * inventory service had to fetch everything every time, calculate
   an offset, and send that back

Instead, we now just ask for 20 UUIDs at a time. This puts less of a
burden on inventory service, removes a URL length limitation, and
overall is cleaner (IMO).

This is a backwards-compatible improvement that can deploy at any time.

### DIFF
--- a/drift/inventory_service_interface.py
+++ b/drift/inventory_service_interface.py
@@ -43,24 +43,29 @@ def _fetch_url(url, auth_header, logger):
     helper to make a single request
     """
     logger.debug("fetching %s" % url)
-    response = requests.get(url, headers=auth_header)
+    with metrics.inventory_service_requests.time():
+        response = requests.get(url, headers=auth_header)
     logger.debug("fetched %s" % url)
     _validate_inventory_response(response, logger)
     return response.json()
 
 
-def _fetch_paginated_url(url, auth_header, logger):
+def _fetch_data(url, auth_header, system_ids, logger):
     """
-    helper to collate a paginated response
-    """
-    results = []
-    page = 1
-    response_json = {'total': 1}
+    fetch system IDs in batches of 40 for given RESTful URL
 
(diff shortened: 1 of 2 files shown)


drift/inventory_service_interface.py
```diff
@@ -49,18 +49,19 @@ def _fetch_url(url, auth_header, logger):
     return response.json()
 
 
-def _fetch_paginated_url(url, auth_header, logger):
+def _fetch_data(url, auth_header, system_ids, logger):
     """
-    helper to collate a paginated response
+    fetch system IDs in batches of 20 for given RESTful URL
     """
+    BATCH_SIZE = 20
```
ehelms: Is this the most optimal value?

beav: I believe 53 was as many as I could get before I hit the URL limit. I chose 20 for human and infra purposes moreso than performance. I did not benchmark various batch sizes, but I was able to get 3K comparisons in 90 sec with batch size of 20.

beav: I added a comment to the code to capture this so it doesn't get lost later

ehelms: Are you able to run the 3k test fairly easily? Would you be up for testing 50 and 100 for batch sizes?

beav: test results for 2K systems:

20 = 2m10s
40 = 1m40s
50 = 1m30s

I changed this to 40 to leave some overhead for URL length limits.

ehelms: Glad to see there was improvements for larger batch sizes. Seems to be trending towards smaller lengths. I am good with sticking to 40 for now.

PR #94, opened 2019-05-22
parse package names differently

beav:
Previously, we used a regex to parse package names. This worked for
most but not all packages.

In this commit, we break the package name into name + VRA. We only
need the VRA as a single string, so there's no need to break it apart
and possibly run into parsing errors.


beav:
credit to @jlsherrill for suggesting this way of doing things

PR #96, opened 2019-05-29
enable cloudwatch log uploading

beav:
This commit enables pushing logs to cloudwatch. It does not disturb
existing stdout logging.

### DIFF
--- a/drift/app.py
+++ b/drift/app.py
@@ -1,7 +1,9 @@
-import connexion
 import logging
 
+import connexion
+
 from drift import config
+from drift.cloudwatch import setup_cw_logging
 from drift.views import v1
 from drift.error import handle_http_error
 from drift.exceptions import HTTPError
@@ -26,6 +28,7 @@ def create_app():
     gunicorn_logger = logging.getLogger('gunicorn.error')
     flask_app.logger.handlers = gunicorn_logger.handlers
     flask_app.logger.setLevel(gunicorn_logger.level)
+    setup_cw_logging(flask_app.logger)
 
     flask_app.register_blueprint(v1.section)
     flask_app.register_error_handler(HTTPError, handle_http_error)
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,8 @@ gunicorn = "*"
 prometheus_client = "*"
 openapi-spec-validator = "==0.2.4"
 bitmath = "*"
+boto3 = "*"
+watchtower = "*"
 
 [requires]
 python_version = "3.6"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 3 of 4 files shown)


drift/app.py
```diff
@@ -1,13 +1,51 @@
...
+        return
+
+    session = Session(
+        aws_access_key_id=key_id,
+        aws_secret_access_key=secret,
+        region_name=os.environ.get("AWS_REGION", "us-east-1"),
+    )
+
+    try:
+        with open("/var/run/secrets/kubernetes.io/serviceaccount/namespace", "r") as f:
```
ehelms: What is this reading?

ehelms:
Does this fail gracefully in development?

drift/app.py
```diff
@@ -1,13 +1,51 @@
...
+from boto3.session import Session
+
 from drift import config
 from drift.views import v1
 from drift.error import handle_http_error
 from drift.exceptions import HTTPError
 from drift.metrics_registry import create_prometheus_registry_dir
 
 
+def _setup_cw_logging(logger):  # pragma: no cover
```
ehelms: I know its a simple method but I'd benefit from this in a separate cloudwatch file of some sort.

drift/app.py
```diff
@@ -1,13 +1,51 @@
...
+    """
+    initialize cloudwatch logging
+
+    from https://github.com/RedHatInsights/cloudwatch-test
+    """
+    key_id = os.environ.get("CW_AWS_ACCESS_KEY_ID")
+    secret = os.environ.get("CW_AWS_SECRET_ACCESS_KEY")
+    if not (key_id and secret):
+        logger.info("CloudWatch logging disabled due to missing access key")
+        return
```
ehelms: Ahh, here is the gracefulness. I missed that. Ignore other comment.

drift/app.py
```diff
@@ -1,13 +1,51 @@
...
+def _setup_cw_logging(logger):  # pragma: no cover
+    """
+    initialize cloudwatch logging
+
+    from https://github.com/RedHatInsights/cloudwatch-test
+    """
+    key_id = os.environ.get("CW_AWS_ACCESS_KEY_ID")
+    secret = os.environ.get("CW_AWS_SECRET_ACCESS_KEY")
+    if not (key_id and secret):
+        logger.info("CloudWatch logging disabled due to missing access key")
```
ehelms: Should we error "in production" if these are not set? We could miss that we are not capturing logs in production which would hurt our ability to debug.

drift/app.py
beav: It is from https://docs.okd.io/latest/dev_guide/service_accounts.html#using-a-service-accounts-credentials-inside-a-container

drift/app.py
beav: I'm a little hesitant to add that only because I don't want to add "this only runs on non-dev deployments" code.

We could add an internal monitoring job to confirm logs are landing maybe?

ehelms: @kylape What have been y'alls approach to this question I have around ensuring logs are landing in production?

PR #97, opened 2019-05-31
Add drift-manifest

ehelms:


beav:
@ehelms I tried `check-manifest.sh` and it didn't fire for my test case. I made the following edit and it worked:

```bash
#!/bin/bash

./scripts/create-manifest.py

changed=`git diff --name-only HEAD^`

if [[ $changed == *"Pipfile.lock"* ]] && [[ $changed != *"drift-manifest"* ]]; then
  echo "Pipfile.lock changed without updating drift-manifest. Run ./scripts/create-manifest.py to update."
  exit 1
else
  exit 0
fi
```

beav:
@ehelms I did not see line 3 when reviewing this, my bad.

ACK looks good, thx

PR #98, opened 2019-05-31
Reduce amount of builds kept and memory limit to default

ehelms:


beav:
can you add a CPU limit as well? The default is 100 or maybe 200 which is a bit high, we could do 25 or 50.

PR #101, opened 2019-06-03
use strings instead of ints for MTU

beav:


ehelms:
This now has conflicts as I merged some other PRs.

PR #103, opened 2019-06-04
Sort systems in fact information by display name

johnsonm325:


drift/info_parser.py
```diff
@@ -106,7 +110,7 @@ def _select_applicable_info(systems_with_profiles):
     return info_comparisons
 
 
-def _parse_profile(system_profile):
+def _parse_profile(system_profile, display_name):
     """
     break complex data structures into more simple structures that can be compared easily.
```
beav: I would add a comment here that display_name is used for sorting but is ultimately stripped out

johnsonm325: But it's not stripped out in this function. Do you still want me to add the comment here?

PR #105, opened 2019-06-10
use black for formatting

beav:
This commit does the following:

 * enable black for syntax checking in travis
 * install black to dev environment
 * make needed python code changes for black compliance

See https://github.com/python/black for more detail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
 script:
   - .travis/check-manifest.sh
   - flake8
+  - black . --check -t py36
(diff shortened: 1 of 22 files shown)


ehelms:
Odd, the manifest check is failing:

```
0.01s$ .travis/check-manifest.sh
.travis/check-manifest.sh: ./scripts/create-manifest.py: /bin/env: bad interpreter: No such file or directory
The command ".travis/check-manifest.sh" exited with 0.
```

PR #106, opened 2019-06-14
Fix shebang in create-manifest

ehelms:


.travis/check-manifest.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 ./scripts/create-manifest.py
+git status
+git diff
```
beav: these two lines are just checking that they both return 0?

ehelms: Oops, no. They were purely debugging to figure out why Travis was acting different. I've removed them.

PR #113, opened 2019-08-02
Kerlescan

beav:


standalone_flask_server.py
```diff
@@ -1,3 +1,3 @@
 from drift.app import create_app
 
-create_app().run(host="0.0.0.0", port=8080)
+create_app().run(host="0.0.0.0", port=8085)
```
ehelms: Why does the port change here?

beav: ah, this snuck in, but the port kept overlapping with baseline service and I kept having to edit one or the other so I could run both when testing.

Since it's already in the PR I think I'll leave it since I kept having to change it locally anyway.

PR #114, opened 2019-08-13
validate baseline UUIDs

beav:


beav:
test instructions:

attempt to fetch a report with one system and one baseline. Instead of a UUID, put a gibberish string or 1234

PR #115, opened 2019-08-13
validate POSTed comparison report via openapi

beav:


beav:
to test:

do a post to create a report. However, instead of system_ids, put "system_id"

PR #123, opened 2019-10-08
Kerlescan bump

beav:


beav:
merging this without review, needed to get CI working

PR #125, opened 2019-10-28
    Display "N/A" as "" if it's from a baseline

beav:
We were previously displaying "N/A" for undefined baseline facts. This
    caused a lot of noise on the comparison table since baselines can have
    lots of undefined facts.
    
    We now display the empty string for baseline "N/A".
    
    Additionally, we strip out "name" as a fact. This was done before but
    regressed.

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -23,11 +23,11 @@ def build_comparisons(systems_with_profiles, baselines):
     )
 
     # remove system metadata that we put into to the comparison earlier
-    stripped_comparisons = [
-        comparison
-        for comparison in fact_comparisons
-        if comparison["name"] not in {"id", "system_profile_exists", "is_baseline"}
-    ]
+    stripped_comparisons = []
+    metadata_fields = {"name", "id", "system_profile_exists", "is_baseline"}
+    for comparison in fact_comparisons:
+        if comparison["name"] not in metadata_fields:
+            stripped_comparisons.append(comparison)
 
     grouped_comparisons = _group_comparisons(stripped_comparisons)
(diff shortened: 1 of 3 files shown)


drift/info_parser.py
```diff
@@ -26,7 +26,8 @@ def build_comparisons(systems_with_profiles, baselines):
     stripped_comparisons = [
         comparison
         for comparison in fact_comparisons
-        if comparison["name"] not in {"id", "system_profile_exists", "is_baseline"}
+        if comparison["name"]
+        not in {"name", "id", "system_profile_exists", "is_baseline"}
```
ehelms: Should these be indented or are they independent lines from the for loop and if block? Man, list comprehensions can be really hard to read

drift/info_parser.py
```diff
@@ -166,17 +167,24 @@ def _create_comparison(systems, info_name):
...
 
+    # change baseline "N/A" to empty string for better display, and remove
+    # fields we added to assist with sorting
+    for sorted_system_id_value in sorted_system_id_values:
+        if (
+            sorted_system_id_value["is_baseline"]
+            and sorted_system_id_value["value"] == "N/A"
+        ):
+            sorted_system_id_value["value"] = ""
+        del sorted_system_id_value["name"]
```
ehelms: Mind an empty line above this? Visually it melds into the if block

drift/info_parser.py
beav: its formatted like this from `black`. Lemme see if i can rewrite it, it's kinda grown over time.

beav:
I rewrote a couple of loops to make them more readable

PR #126, opened 2019-10-31
add version to API

beav:
This commit adds a version variable that is served from `/version`.
This can be used by python-semantic-release for versioning, which will
be set up in a later commit.


drift/app_config.py
```diff
@@ -1,8 +1,17 @@
 import os
 
+drift_app_version = "0.0.1"
```
ehelms: I tend to expect the version to be more in a setup.py file or VERSION file. This does depend on what the intended workflow is for bumping the version.

PR #129, opened 2019-12-05
use "unique" check for certain network facts

beav:
Previously, we were raising an alert if two systems had different IP
addresses or HW addresses.

Instead, only raise an alert if we don't have unique addresses across
all systems compared.

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -173,6 +173,17 @@ def _create_comparison(systems, info_name):
     elif len(system_values) == 1:
         info_comparison = COMPARISON_SAME
 
+    # override comparison logic for certain fact names
+    if _is_unique_rec_name(info_name):
+        system_values_with_duplicates = [system["value"] for system in system_id_values]
+        if len(system_values) == len(system_values_with_duplicates):
+            info_comparison = COMPARISON_SAME
+        else:
+            info_comparison = COMPARISON_DIFFERENT
+
+    if _is_no_rec_name(info_name):
+        info_comparison = COMPARISON_INCOMPLETE_DATA
+
     # change baseline "N/A" to empty string for better display, and remove
     # fields we added to assist with sorting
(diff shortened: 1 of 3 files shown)


tests/test_v1_api.py
```diff
@@ -95,6 +95,15 @@ def test_comparison_report_api(self, mock_fetch_systems):
...
         self.assertEqual(response.status_code, 200)
+        comparisons = json.loads(response.data)
+        # we hard-code the index lookup since we know the fixture layout
+        network_comparisons = comparisons["facts"][12]["comparisons"]
+        ipv4_comparison = [
+            comparison
+            for comparison in network_comparisons
+            if comparison["name"] == "eth99.ipv4_addresses"
+        ][0]
+        self.assertEqual(ipv4_comparison["state"], "DIFFERENT")
```
ehelms: Could we add individual tests for each unique fact to ensure none of them regress?

drift/info_parser.py
```diff
@@ -189,6 +203,23 @@ def _create_comparison(systems, info_name):
...
+    if not info_name.startswith("network_interfaces"):
+        return
+    elif info_name.startswith("network_interfaces.lo."):
+        return
+    elif info_name.endswith(".ipv4_addresses"):
+        return True
+    elif info_name.endswith(".ipv6_addresses"):
+        return True
+    elif info_name.endswith(".mac_address"):
+        return True
```
ehelms: We need the fqdn included as well as a unique recommendation

ehelms: Thoughts on a dictionary stored either in a separate python file or json/yaml object that maps property to its operator? And default to 'same' if no entry.

beav: I can split it out into an array which may help with readability

beav:
@ehelms updated

PR #130, opened 2019-12-05
feature: new comparison on fqdn and network facts

beav:
Previously, we checked all facts to see if they were the same.
However, this does not make sense for the following facts:

* ipv4 address
* ipv6 address
* fqdn
* last boot time

We now check that the first three are unique, and we do not do any
comparison on last boot time.


beav:
@ehelms i had a short-circuit before to skip logic that was over-aggressive and skipped fqdn

PR #137, opened 2020-01-21
feature: rename PIT to HSP

beav:
Previously, we used PIT in some places and HSP in others. This commit
gets rid of as many references to PIT as possible. There are a couple
of remaining references in kerlescan that still need to be removed.

This commit also breaks historical system profiles out as a peer to
`system` and `baseline`. Previously, we lumped them in with `system`s
for proof of concept purposes.


drift/hsp_service_interface.py
```diff
@@ -6,22 +6,22 @@
 from kerlescan.constants import AUTH_HEADER_NAME, PIT_SVC_ENDPOINT
 
 
-def fetch_pits(pit_ids, service_auth_key, logger):
+def fetch_hsps(hsp_ids, service_auth_key, logger):
```
ehelms: Long names suck but `hsps` is a hard one for me to translate mentally everytime I read a line of code. Could we enumerate it fully or shorten it to something more like `historical_profiles` or `historical_sys_profiles` ?

beav: i had `historical_profiles` but wanted to leave the S in there. I'll go w/ `historical_sys_profiles`

drift/historical_sys_profile_service_interface.py
```diff
@@ -0,0 +1,27 @@
+from urllib.parse import urljoin
+
+from drift import metrics
+from kerlescan import config
+from kerlescan.service_interface import fetch_data
+from kerlescan.constants import AUTH_HEADER_NAME, PIT_SVC_ENDPOINT
```
ehelms: For consistency, should PIT_SVC_ENDPOINT be renamed?

ehelms:
This will need a corresponding UI update, correct?

drift/historical_sys_profile_service_interface.py
beav: it does. I'll rename it in a different PR so I can update kerlescan at the same time. Trying to not break too many things at once :)

beav:
> This will need a corresponding UI update, correct?

correct. The plan right now is to merge this (and break CI beta for the feature), then @johnsonm325 can make the UI changes.  We wouldn't merge this one until @johnsonm325 is ready to work on it, to keep the breakage window small.

I'll work today on making the local dev environment easier to stand up. I think the plan above is OK for now though, given its CI beta, and this change only breaks the new feature.

PR #140, opened 2020-02-04
fix: rename PIT to HSP

beav:
We are working to use "HSP" instead of "PIT" to avoid having two names
for the same thing.


beav:
this needs https://github.com/beav/kerlescan/pull/9

PR #143, opened 2020-02-06
fix: add inventory_id to HSP metadata

beav:


drift/info_parser.py
```diff
@@ -282,4 +282,5 @@ def _historical_sys_profile_mapping(historical_sys_profile):
         "id": historical_sys_profile["id"],
         "display_name": historical_sys_profile["display_name"],
         "updated": historical_sys_profile["updated"],
+        "inventory_id": historical_sys_profile["inventory_id"],
```
ehelms: I am leaning more towards `system_id` since we use the term `system` and `system_id` everywhere else that I could find in our APIs.

beav: i'll fix it up here. I'll change it later on in the HSP model obj

PR #147, opened 2020-03-12
Merge to stable

beav:


beav:
These commits level-set master with stable. The diff between the two should be the same except for some library minor version updates in Pipfile.lock.

PR #152, opened 2020-03-24
reference ID

beav:


drift/info_parser.py
```diff
@@ -217,6 +222,18 @@ def _create_comparison(systems, info_name):
...
+        # pull the reference_value for this comparison
+        reference_value = None
+        for values in sorted_system_id_values:
+            if values["id"] == reference_id:
+                reference_value = values["value"]
+
+        for values in sorted_system_id_values:
+            values["meets_reference_expectations"] = True
+            if values["value"] != reference_value:
+                values["meets_reference_expectations"] = False
```
ehelms: I may have missed a discussion. I was still thinking we were going with mimicking the `state` concept here so that we can support more than true/false down the road.

beav: you didnt miss anything, I just goofed and left it like this :smile_cat: 

would "expected" and "unexpected" work?

ehelms: Jerome and I landed on using `state` and the same states we do for comparison. This aligns with that view of state and provides flexibility. What do you think of that idea?

beav: that works for me. So basically the state for the array of comparisons is the "overall" state and then each dict could have an override state.

PR #155, opened 2020-04-01
fix: return csv output for hsps in comparison request

CodeHeeler:
The backend failed to return csv output for hsps in system comparison
requests.  It now returns the hsps csv output.

### DIFF
--- a/drift/views/v1.py
+++ b/drift/views/v1.py
@@ -75,7 +75,7 @@ def _csvify(comparisons):
     fieldnames = ["name", "state"]
     record_ids = []
     system_names = {}
-    # add baselines to the CSV dict, then systems
+    # add baselines to the CSV dict, then systems, then historical system profiles
     for baseline in comparisons["baselines"]:
         record_ids.append(baseline["id"])
         system_names[baseline["id"]] = baseline["display_name"]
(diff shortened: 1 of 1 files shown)


CodeHeeler:
I opted to have my added code accommodate existing decisions, but note that the only place until now where we write out 'system' in historical system profiles in the code is in the 'build_comparisons' method here:
https://github.com/RedHatInsights/drift-backend/blob/7210747ad729164021a683f1e6c5ba8b9e6baf88/drift/info_parser.py#L57

and also in this yaml file:
https://github.com/RedHatInsights/drift-backend/blob/master/drift/openapi/api.spec.yaml#L188

It might be more pythonic and consistent to change this to historical_sys_profiles as it is everywhere else so that it's more predictable what key will be expected when referencing these.

Happy to make this change as part of this PR if it is agreed upon, or open a card for this issue and make another quick PR to handle this change, just let me know what you all think.

ehelms:
Since `historical_system_profiles` is user facing in the API response, I would like to keep the verbosity of `system` in that aspect. If I understand correctly, that is what this change would mean and also affect the UI. If the internal python code switches to use `_system_` instead of an abbreviated `_sys_` for consistency that is good by me.

CodeHeeler:
Ok, in that case, this PR is good to go.  I'll discuss the internal python case with others and as that would be a larger refactor, we'll create a card if we intend to tackle that work.  Thanks!

PR #162, opened 2020-04-27
fix: check hsp ids exist

CodeHeeler:


beav:
This change works, but it would probably be better if `historical_sys_profile_service_interface` lived in kerlescan, similar to `inventory_service_interface`.

Do you have a preference between moving it over now, or keeping this PR as-is and doing that change later on?

CodeHeeler:
I have a mild preference for just getting more of these fixes knocked out now that folks are using it, and handling the refactors as their own separate cards.

beav:
This works for me when I tested :+1: 

Can you move the logic in `_ensure_correct_system_count` into a small method in `historical_sys_profile_interface.py`? It creates a small amount of code duplication between drift and kerlescan but that way you aren't importing a private method. We can remove it when we do the refactor.

PR #163, opened 2020-05-06
bump kerlescan to 0.16

beav:
bump to 0.16 for cloudwatch conf update


beav:
merging since its just a version bump of code thats already acked

PR #166, opened 2020-05-29
404 if hsp profile_id non-existent

CodeHeeler:


beav:
I see the refactor to move the hsp fetch out of the other method's params (yay) but I don't see any other changes. Am I missing something, or is the fix in kerlescan?

CodeHeeler:
This is another one of those that the kerlescan change fixed I think. I verified that it works currently.

PR #169, opened 2020-06-09
bump kerlescan to 0.26

beav:
I didn't tag the latest for 0.25, oops. 0.26 is correct.


beav:
merging, its a fix to something i broke before during lib tagging.

PR #171, opened 2020-06-11
boto3 bump

beav:


beav:
no code change, version bump because the older version is having sha issues. merging.

PR #172, opened 2020-06-15
only enable "unique" checks for multi-system

beav:
Previously, we checked that certain facts like FQDN were unique across
systems. This makes sense for systems but not baselines or HSPs.

Instead, only use the uniqueness check if multiple systems are present.


beav:
The unit tests in this area are pretty weak. I will update them in a future commit after we ensure this behavior is what we want.

PR #174, opened 2020-07-06
add report generation test

beav:
This commit adds a report generation test that gets run after unit
tests. It also fixes a small bug related to ordering of items inside
of a grouped comparison.


generate_report.py
```diff
@@ -0,0 +1,46 @@
...
+systems, profiles, baselines, hsps = [], [], [], []
+with open("tests/reports/inputs/inventory/report1-systems.json", "r") as myfile:
+    systems = json.loads(myfile.read())["results"]
+
+with open("tests/reports/inputs/inventory/report1-system_profiles.json", "r") as myfile:
+    profiles = json.loads(myfile.read())["results"]
+
+systems_with_profiles = interleave_systems_and_profiles(systems, profiles)
+
+with open("tests/reports/inputs/inventory/report1-system_profiles.json", "r") as myfile:
```
CodeHeeler, line 21: This is a duplicate of lines 16-17 above.

generate_report.py
```diff
@@ -0,0 +1,46 @@
...
+print("Testing report generation...")
+# TODO: remove need to initialize flask app
+app = create_app()
+app.app.app_context().push()
+
+systems, profiles, baselines, hsps = [], [], [], []
+with open("tests/reports/inputs/inventory/report1-systems.json", "r") as myfile:
+    systems = json.loads(myfile.read())["results"]
+
+with open("tests/reports/inputs/inventory/report1-system_profiles.json", "r") as myfile:
```
CodeHeeler, line 16: Can you just use the same name (myfile) over and over while it's still open from the previous file like this?

generate_report.py
```diff
@@ -0,0 +1,46 @@
...
+    hsps = json.loads(myfile.read())["data"]
+
+comparisons = info_parser.build_comparisons(
+    systems_with_profiles, baselines, hsps, None
+)
+
+
+# to test generated report against expected data:
+with open("tests/reports/outputs/report1.json", "r") as myfile:
+    result = json.dumps(comparisons, indent=4, sort_keys=True).splitlines()
```
CodeHeeler, line 37: You used sort_keys=True here which will alphabetize them but you do not alphabetize the 'expected' json below so this comparison will probably fail.

generate_report.py
beav: oops!

generate_report.py
beav: it closes once it leaves the "with" statement due to python magic :mage_woman: . Basically "with" will create a context where the file is open, and automatically close it once we leave the code block.

If I didn't use the with statement, I'd have to do something like this:

```python
try:
  myfile = open("tests/reports/inputs/inventory/report1-systems.json", "r")
  systems = json.loads(myfile.read())["results"]
except:
  # handle exceptions here
finally:
  myfile.close()

https://docs.python.org/3.8/reference/compound_stmts.html#with

generate_report.py
beav: it is alphabetized when saving from the `sort_keys=True` in the section below that writes the report. I guess I can add a command-line option to write a new report JSON instead of doing it the current way of commenting/uncommenting, i'll do that tomorrow.

generate_report.py
beav: I cleaned this up, marking as resolved

PR #177, opened 2020-07-20
update manifest format

CodeHeeler:


scripts/create-manifest.py
```diff
@@ -11,9 +11,9 @@
     for name, value in sorted(lockfile["default"].items()):
         if "version" in value:
             version = value["version"].replace("=", "")
-            manifest.write("drift/python:3.6=%s:%s\n" % (name, version))
+            manifest.write("services-drift:drift/python:3.6=%s:%s\n" % (name, version))
```
beav, line 14: Based on feedback I received today, please do this instead (basically remove the sub-component):

`services-drift/python`

Thanks!

beav: actually hold off for now, still looking.

beav: It looks like the correct format would be something like this:

```
             manifest.write(f"service-drift:{epoch}/{name}.{version}\n")
```

where `epoch` is the epoch. You can get the epoch like so:
```python
import time
from datetime import datetime

dt = datetime.today()
epoch = time.mktime(dt.timetuple())

print(f"epoch is {epoch}")

```

Just make sure you pull the epoch once before the loop starts, otherwise it will increment as the loop gets executed :)

PR #179, opened 2020-07-20
add validation for reference_id

CodeHeeler:


beav:
It looks unhappy about getting None as a UUID:

```
  File "/home/travis/virtualenv/python3.6.7/src/kerlescan/kerlescan/view_helpers.py", line 144, in validate_uuids

    system_id.lower(),

AttributeError: 'NoneType' object has no attribute 'lower'
```

PR #182, opened 2020-07-29
correct state for parent of subfact

CodeHeeler:


beav:
the change looks OK. it looks like you need to do another saved report regen to update the expected test results.

PR #199, opened 2020-09-24
fqdn incomplete for n/a, account for baselines

CodeHeeler:


drift/info_parser.py
```diff
@@ -218,13 +218,20 @@ def _create_comparison(systems, info_name, reference_id, system_count):
...
         info_comparison = COMPARISON_SAME
 
+    baseline_count = 0
+    for system_id_value in sorted_system_id_values:
+        if system_id_value["is_baseline"]:
+            baseline_count += 1
+
     # override comparison logic for certain fact names
-    if _is_unique_rec_name(info_name) and system_count > 1:
+    if _is_unique_rec_name(info_name) and (system_count - baseline_count) > 1:
```
beav, line 227: can you add a comment that explains what this is checking for? if i came across this code later I wouldn't remember why it was doing this

drift/info_parser.py
```diff
@@ -218,13 +218,20 @@ def _create_comparison(systems, info_name, reference_id, system_count):
...
         system_values_with_duplicates = [system["value"] for system in system_id_values]
         if len(system_values) == len(system_values_with_duplicates):
             info_comparison = COMPARISON_SAME
         else:
             info_comparison = COMPARISON_DIFFERENT
+        if "N/A" in system_values:
```
beav, line 233: should this be

```python
if:
  blah
elif "N/A" in system_values:
    info_comparison = COMPARISON_INCOMPLETE_DATA
else:
    info_comparison = COMPARISON_DIFFERENT
```
?

tests/reports/outputs/report1.json
```diff
@@ -57769,7 +57769,7 @@
             "comparisons": [
                 {
                     "name": "docker0.ipv4_addresses",
-                    "state": "DIFFERENT",
+                    "state": "INCOMPLETE_DATA",
```
beav, line 57772: are the changes here expected? or should the status only be altered for FQDN?

drift/info_parser.py
CodeHeeler: I saw this more as an override.  Could do as you say and put incomplete in the first 'if' position.  Is there a reason that is better?

tests/reports/outputs/report1.json
CodeHeeler: the changes here are expected because fqdn is treated the same as *_addresses, all are considered unique rec names, as far as I understand it.

drift/info_parser.py
beav: only for readability and my personal preference. If you think your way is more readable, go with that. A comment may be helpful in either case

tests/reports/outputs/report1.json
CodeHeeler: To clarify, with recent work, we are changing behaviour for statuses to be Incomplete for any "N/A" data, and that's not specific to fqdn, it's just one of many places the change was made and needed some extra logic, if I recall correctly.

drift/info_parser.py
CodeHeeler: Ok, added a comment and shuffled for readability, thanks!

beav (approved):

PR #206, opened 2020-10-22
only include sap facts if sap system present

CodeHeeler:


drift/info_parser.py
```diff
@@ -191,6 +192,13 @@ def _create_comparison(systems, info_name, reference_id, system_count):
 
     """
     # TODO: this method is messy and could be refactored
+    sap_present = False
+    for system in systems:
+        if "sap_system" in system and system["sap_system"]:
+            sap_present = True
+    if not sap_present:
+        return
+
```
eherget, line 201: This looks like it will skip comparing all facts, so ultimately only sap systems will have any facts compared. Am I reading this wrong? I have not tried to run this code yet.

CodeHeeler: good catch, removed one too many if statements in a rewrite.  fixed!

eherget (approved):
LGTM

PR #208, opened 2020-10-27
Catch inventory RBAC error

CodeHeeler:
Dependent on: https://github.com/RedHatInsights/kerlescan/pull/1

Will need to update to include kerlescan changes once tagged.


eherget (approved):
This is straight-forward change and looks good.  Approving.

PR #209, opened 2020-10-27
sap bool is true or undef, not false

CodeHeeler:


eherget (approved):
Looks like it should work.  Its easier to test this in CI than to spin up my local env.

PR #216, opened 2021-01-26
DRFT-39 Add multivalue handling to comparisons

CodeHeeler:
Ready for review!

Will need to update this PR to point to a newly tagged kerlescan version before merging.

This adds multivalue facts comparisons logic.

There were not large blocks deleted as it seems to indicate in the Files Changed, but the combo of my indenting a large section and then rebasing with the obfuscated values work really confused git.


CodeHeeler:
Fixed two small issues eherget found in initial review.  Discussing with micjohns now, then will push up a rebase and this should be ready for more in depth testing.

drift/info_parser.py
```diff
@@ -255,69 +255,150 @@ def _create_comparison(systems, info_name, reference_id, system_count):
...
+                    row += 1
+            sorted_system_id_values[value_count]["value"] = column
+            value_count += 1
+
+        # need to account for states compared to a reference if selected
+        #for system in sorted_system_id_values:
+        #    system["state"] = reference_comparison
+
+
+        row_count = len(multivalue_comparisons)
```
eherget, line 384: This is redundant.  row_count value is set in line 350 and doesn't get changed after.

eherget (approved):
lgtm

PR #217, opened 2021-01-26
DRFT-124 Fact is in the SAME state when no values left for comparison

petrblaho:
https://issues.redhat.com/browse/DRFT-124


CodeHeeler (approved):
Ah, I see, so it was trying to check a zero.  Thanks for the fix!

PR #226, opened 2021-02-03
[DRFT-155] Change No Data to empty string for multivalue

CodeHeeler:
Make multivalue facts display the same way as the rest of the comparison report with empty cells in stead of "No Data".

Feel free to discuss here any special cases such as Tags and how you'd like them to be addressed, or we can open that as a bug/separate ticket, either is fine with me.

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -366,7 +366,7 @@ def _create_comparison(systems, info_name, reference_id, system_count):
                     if sorted_all_value_options[row] == value:
                         column.append(value)
                     else:
-                        column.append(None)
+                        column.append("")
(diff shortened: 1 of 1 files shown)


eherget (commented):
Just so I understand, its the UI that put "No data" in the cell if the corresponding value is None.  So this change will prevent the UI from putting anything in the cell since it now has a real value of <empty string>?

CodeHeeler:
Exactly.

eherget (approved):
Ok.  I'll approve this one.
The tags stuff setting "(no value)" happens in kerlescan where we process the system profile.  This happens later, when we do a comparison across one or more profiles already processed by kerlescan.

PR #227, opened 2021-02-04
[DRFT-157][DRFT-158] Fix multivalue schema, add is_obfuscated

CodeHeeler:


eherget (approved):
This appears to be alright.  Have we run it through some validator, or more precisely, run various comparison report outputs through a validator that uses this schema?

CodeHeeler:
Yes, I ran a java validate locally and then using my local drift-dev-setup env ran the systems through, corrected any errors for both multivalue and obfuscation and it completes fine now here.

PR #228, opened 2021-02-10
Update handling of empty values

CodeHeeler:
Here's an initial PR for discussion; we can clean it up further once everyone's on the same page.

--Updated regular facts for systems to be empty string instead of `N/A` to be consistent with baselines and multivalue facts
--Removed `name` and `is_baseline` from systems for multivalue facts as they were fields added to assist with sorting and removed at the end of the normal fact logic but I failed to replicate that for multivalues before now
--Removed `is_obfuscated` for multivalues for now as there is currently no overlap and it is populated with `None` instead of True or False, so is not needed in the json

Do we want to remove `is_obfuscated` from the final comparison report json when the value is False for normal facts also or only when the value is None?  See discussion on DRFT-158.

Note: lowered coverage report for this to work for now; increasing coverage and adding tests is covered by DRFT-146 which is the next priority.

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -267,7 +267,7 @@ def _create_comparison(systems, info_name, reference_id, system_count):
         }
 
         if "N/A" in system_values:  # one or more values are missing
-            info_comparison = COMPARISON_INCOMPLETE_DATA
+            info_comparison = COMPARISON_DIFFERENT
         elif (
             len(system_values) <= 1
         ):  # when there is only one or zero non-obfuscated values left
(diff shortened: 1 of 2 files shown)


eherget:
My understanding of how you implemented this is that the "special values" like "N/A" are removed just before the comparison report output is returned to the calling client. My preference is for not using special values for a couple reasons.  Currently we still use them internally as, IIRC, baseline creation chokes on facts that have empty or no value. This PR is a good step to removing special values externally and helps clean up the UI, IMO.

As for is_obfuscated - my preference would be to only include it on fact values that are obfuscated, if for no other reason than to reduce the size of the output.  However, we're using a text-based output format, so size of the output is clearly not the driving factor for how it was implemented.  ;-)

eherget (approved):
This looks good.  I'll approve it.  But let's open a card to increase unit test coverage as I don't like bumping it down as we ahve been.

CodeHeeler:
@eherget same, already done before the previous PR and it's my next task: https://issues.redhat.com/browse/DRFT-146

PR #229, opened 2021-02-11
Remove is_obfuscated if False

CodeHeeler:


drift/info_parser.py
```diff
@@ -305,6 +305,8 @@ def _create_comparison(systems, info_name, reference_id, system_count):
 
             del system_id_value["name"]
             del system_id_value["is_baseline"]
+            if not system_id_value["is_obfuscated"]:
+                del system_id_value["is_obfuscated"]
```
beav, line 309: You can do this in one line like so:

```python
system_id_value.pop("is_obfuscated", None)
```

For example:

```
>>> x = {"a": 1, "b": 2, "c": 3}
>>> x
{'a': 1, 'b': 2, 'c': 3}
>>> x.pop('a')
1
>>> x
{'b': 2, 'c': 3}
>>> x.pop('a')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'a'
>>> x.pop('a', None)
>>> x
{'b': 2, 'c': 3}
>>> 
```

beav: If you end up doing it this way, you may want to change the other two `del` also just to make them consistent

CodeHeeler: Does that cover both cases, is_obfuscated=None and is_obfuscated=False?

Also, is there any benefit to doing it this way when we don't need the returned value `pop` provides?  Is it somehow still more efficient than `del`?

eherget: I think there might be a misunderstanding here.  The purpose of the 2 highlighted lines is to remove "is_obfuscated" if it is False, but to leave it in place if it is True.  I think beav may be thinking the goal was to simply remove "is_obfuscated" if it exists, regardless of its value.
I think we know that "is_obfuscated" always exists, so we don't need to check for, or handle the case where it does not exist. The same is true of "name" and "is_baseline".

eherget (approved):
LGTM

PR #230, opened 2021-02-11
If all values compared are empty, state is SAME

CodeHeeler:


eherget (commented):
Your change sort of duplicates what lines 269-274 do, with the exception of when the only value in system_values is "N/A".  Even if this is re-worked so its COMPARISON_SAME when "N/A" is the only value in system_values, possibly code after line 274 changes it to something else?  Is there a way to avoid such confusion of seeing what looks like the same effort to set COMPARISON_SAME when len(system_values) ==1 in two different places?

petrblaho:
https://issues.redhat.com/browse/DRFT-194

petrblaho:
@eherget I agree with you that seeing the same kind of assignment in multiple places is confusing but as you wrote the value can be changes after line 274.

So I guess that now it has to do with priorities of how to set the comparison result. I think that @CodeHeeler wanted to be sure that it was set this way even in case the code after 274 changed the value.

CodeHeeler:
I like @petrblaho 's proposed rewrite of comparisons code using a state machine, and we'll pull that into another sprint.  Can this move forward for now so that we unblock all of the waiting features in ci to move to qa? What do you think @eherget

eherget (approved):
I agree with you both.  Approved.
DRFT-194 will address the potential confusion I commented on.

PR #238, opened 2021-03-12
[DRFT-176] Add a short_circuit to comparisons

CodeHeeler:
When a system checks in and we create a new hsp, we need to see
if anything has changed, and if so, alert Notifications.
To do this, we need to run a comparison but we only care if any
difference exists at all, not about the whole report.

This is probably the safest PR to merge first as it shouldn't change anything since short_circuit defaults to False, but we can't test it until after the DB changes, so maybe we should hold it until after that.  Once our DB change to models in baselines PR is ready to go, then we can add the changes from the hsp/kerlescan PRs and test this out.

Note a lot of this could be simplified by importing the comparison methods from drift directly into the hsp archiver and bypassing the api but that would closely couple the apps, so by adding the interactions with the api, we can continue to use kerlescan as the only link to worry about versioning.  If we revisit this architecture and either move the core of the comparison into kerlescan or move kerlescan back into the apps, we can also rethink this code.

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management


petrblaho (approved):
I think this would work.

What I personally don't like is that we change the pool of return types for these methods from `dict` to `dict or bool` but I can live with that.

petrblaho (changes requested):
I just found out the we call `build_comparison` at https://github.com/RedHatInsights/drift-backend/blob/master/generate_report.py#L67..

So we need to modify for that.

CodeHeeler:
> I think this would work.
> 
> What I personally don't like is that we change the pool of return types for these methods from `dict` to `dict or bool` but I can live with that.

I agree, this is not ideal, but I don't want to duplicate the whole method elsewhere either.  Maybe when we have our refactor discussion with a logic flow chart for comparisons, we can find a better solution.

CodeHeeler:
> I just found out the we call `build_comparison` at https://github.com/RedHatInsights/drift-backend/blob/master/generate_report.py#L67..
> 
> So we need to modify for that.

Oooh, excellent catch, thanks!

petrblaho:
> > I think this would work.
> > What I personally don't like is that we change the pool of return types for these methods from `dict` to `dict or bool` but I can live with that.
> 
> I agree, this is not ideal, but I don't want to duplicate the whole method elsewhere either. Maybe when we have our refactor discussion with a logic flow chart for comparisons, we can find a better solution.

I see it the same way here...

CodeHeeler:
> > > I think this would work.
> > > What I personally don't like is that we change the pool of return types for these methods from `dict` to `dict or bool` but I can live with that.
> > 
> > 
> > I agree, this is not ideal, but I don't want to duplicate the whole method elsewhere either. Maybe when we have our refactor discussion with a logic flow chart for comparisons, we can find a better solution.
> 
> I see it the same way here...

Alternatively, if we're not worried about performance down to the level of short-circuiting from within the comparison method itself, we could let it complete and return the dict for that individual fact, then check the state in the dict for that fact to see if it's DIFFERENT only there.  It would keep us from adding complications to the already messy comparison itself, and it wouldn't alter the potential return values for that method (though it still does in the calling methods).  Then, if we really wanted consistent return values, we could pull the remaining code out to helper functions and have a different wrapper for calling the get if we're using it only for this notification purpose.  Thoughts?

petrblaho:
> > > > I think this would work.
> > > > What I personally don't like is that we change the pool of return types for these methods from `dict` to `dict or bool` but I can live with that.
> > > 
> > > 
> > > I agree, this is not ideal, but I don't want to duplicate the whole method elsewhere either. Maybe when we have our refactor discussion with a logic flow chart for comparisons, we can find a better solution.
> > 
> > 
> > I see it the same way here...
> 
> Alternatively, if we're not worried about performance down to the level of short-circuiting from within the comparison method itself, we could let it complete and return the dict for that individual fact, then check the state in the dict for that fact to see if it's DIFFERENT only there. It would keep us from adding complications to the already messy comparison itself, and it wouldn't alter the potential return values for that method (though it still does in the calling methods). Then, if we really wanted consistent return values, we could pull the remaining code out to helper functions and have a different wrapper for calling the get if we're using it only for this notification purpose. Thoughts?

Yeah, I like the idea to just keep it the same and check for DIFFERENT presence. Less code changes for now.

CodeHeeler:
Ok, I've reverted all changes to the _create_comparison method since that's processing only one fact and we don't need to worry about fine tuning at that level now, so it will return the comparison for the one fact, then the calling methods check the state and short circuit back out to notify for the mvp (once we get it hooked up and working all the way through, the next goal will be to return the comparison report as well, but I don't want to take more time on it for now since our cleanup sprint is likely to address refactoring the entire comparison).

I've also made the change so that only facts present in the baseline will be checked since only those matter for the notification.

petrblaho (approved):
LGTM!

PR #248, opened 2021-04-20
Add short_circuit to spec for comparison

CodeHeeler:

## Secure Coding Checklist
- [x] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -39,13 +39,8 @@ def build_comparisons(
         reference_id,
         short_circuit,
     )
-    if short_circuit:
-        if fact_comparisons:
-            return (
-                True  # do we want to pass another value for case "nothing different"?
-            )
-        else:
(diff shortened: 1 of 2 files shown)


CodeHeeler:
@petrblaho So here's a start on the schema update but I think I need to update the responses to what qualifies as a 200 still to include what we return when short_circuit is True.

Also, it occurs to me this may also need an internal api? It's likely a "both or neither" where we could make both public or both private, but I'm open to what others think.  I've started on that on a separate branch if we decide that's what we want.

petrblaho (approved):

PR #249, opened 2021-04-23
[DRFT-336] Update docs to match notifications behaviour

CodeHeeler:
We're returning a comparison report with an added key on if we
need to notify rather than having short circuit return only a
bool, so updated docstrings to reflect this behaviour.
Also removed a check that should be handed by the try except instead

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -26,11 +26,11 @@ def build_comparisons(
     given a list of system profile dicts and fact namespace, return a dict of
     comparisons, along with a dict of system data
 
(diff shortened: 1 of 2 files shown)


petrblaho (approved):
LGTM!

PR #269, opened 2021-06-29
[DRFT-616] Files modified to be consistent with the new pre-commit hook

fellipeh:
This PR include all modified files by pre-commit hook,

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management
- [x] General Coding Practices


documentation/api.html
```diff
@@ -1458,12 +1458,12 @@
   opacity: 0.5;
 }
 </style>
-    
+
```
sthirugn, line 1461: this file has lots of changes, I guess we need to trust that the tool would have done the right thing? :)

sthirugn:
lgtm once the jenkins job is resolved. Thank you @fellipeh

documentation/api.html
CodeHeeler: Yeah, but then why has this file that hasn't been touched in ages not been fixed already since we run both black and flake8 with every PR?

sthirugn:
Right now, the commit message shows:

```
This PR includes:

 - Files modified to be consistent with the new pre-commit hook
 - flake8 hook (missed on the Pr regarding the pre-commit hooks)
 - changes pre-commit-config files to be consistent with the project CI
 
  ```

Can you change it to something like the following: (to have a clean commit history while doing `git log`?

```
[DRFT-616] Modify files to be consistent with the new pre-commit hook

Changes include:
 - Files modified to be consistent with the new pre-commit hook
 - flake8 hook (missed on the Pr regarding the pre-commit hooks)
 - changes pre-commit-config files to be consistent with the project CI
 
  ```

Note:
- The commit messages prefer active voice, so `modify files` versus `Modified files`

sthirugn (approved):
Looks great, thanks.  

I think @CodeHeeler is still reviewing.

documentation/api.html
sthirugn: @CodeHeeler good question, I am not sure why this file `api.html` was not caught before.  @fellipeh any thoughts?

drift/app.py
```diff
@@ -21,12 +21,8 @@ def create_app():
     """
     app_name = os.getenv("APP_NAME", "drift")
     openapi_args = {"path_prefix": config.path_prefix, "app_name": app_name}
-    connexion_app = connexion.App(
-        __name__, specification_dir="openapi/", arguments=openapi_args
```
vkrizan, line 25: I'd advise to keep the line limit to the previous value that was used to limit number of changes and use what you already are used to. (if these kind of changes were due to the line limit increase)

documentation/api.html
vkrizan: It cleared out whitespace on empty lines. It should be harmless.

drift/app.py
fellipeh: The line limit is set to 100, that's the reason of this change.

Of course we can move and set to 120 col, several project use that (NumPy, SciPy, Pandas) 
Just for note:
- PEP says 79  (https://www.python.org/dev/peps/pep-0008/#maximum-line-length())
- Black set to default 88  (https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html?highlight=line-length#id2)
- Django set to 119 for the code and 79 for the doc string (https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style)

CodeHeeler: I tend to prefer the longer the better :)
I don't like having to jump through hoops to try to force things onto shorter lines.  On the other hand, we don't want to be too far from standard.  I'm used to using black now to figure it out I guess. 79 is too short, in my opinion.  I'd be ok with keeping to the previous value of 100, but I'm open to expanding it if others prefer longer.

sthirugn: +1 to the old value 100

fellipeh: ack, let stay with 100 (this PR change to it)

PR #275, opened 2021-07-29
 DRFT-710: Create admin endpoint on drift-backend

fellipeh:
This PR adds admin endpoint for stage test purpose

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/drift/app.py
+++ b/drift/app.py
@@ -24,6 +24,7 @@ def create_app():
     connexion_app = connexion.App(__name__, specification_dir="openapi/", arguments=openapi_args)
     connexion_app.add_api("api.spec.yaml", validate_responses=True, strict_validation=True)
     connexion_app.add_api("mgmt_api.spec.yaml")
+    connexion_app.add_api("admin_api.spec.yaml", validate_responses=True, strict_validation=True)
     flask_app = connexion_app.app
 
     create_prometheus_registry_dir()
--- a/drift/openapi/admin_api.spec.yaml
+++ b/drift/openapi/admin_api.spec.yaml
@@ -1,0 +1,36 @@
+---
+openapi: 3.0.1
+info:
+  version: "0.1"
+  title: Drift Backend Service Admin Plane
+  description: Admin API for Drift Service
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: "{{ path_prefix }}{{ app_name }}/admin"
+
+
+paths:
+  /status:
+    get:
+      summary: admin data call
+      operationId: drift.views.admin.status
+      tags:
+        - status
+      responses:
+        '200':
+          description: "A small JSON indicating some information about application"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Status"
+components:
+  schemas:
+    Status:
+      required:
+        - status
+      properties:
+        status:
+          type: string
--- a/drift/views/admin.py
+++ b/drift/views/admin.py
@@ -1,0 +1,24 @@
+from flask import Blueprint, jsonify
+
+from drift.version import app_version
+
+
+section = Blueprint("internal_admin", __name__)
+
+
+@section.before_app_request
+def status():
+    return jsonify(
+        {
+            "drift_version": app_version,
+            "status": "Warning! This is a test data.",
+            "totalBaselinesCount": 541,
+            "customerIdsCount": 279,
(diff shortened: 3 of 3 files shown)


drift/views/admin.py
```diff
@@ -0,0 +1,24 @@
...
+
+section = Blueprint("internal_admin", __name__)
+
+
+@section.before_app_request
+def status():
+    """Return"""
+    return jsonify(
+        {
+            "drift_version": app_version,
```
sthirugn, line 13: can you indicate somewhere that this is test data?

drift/app.py
```diff
@@ -24,6 +24,7 @@ def create_app():
     connexion_app = connexion.App(__name__, specification_dir="openapi/", arguments=openapi_args)
     connexion_app.add_api("api.spec.yaml", validate_responses=True, strict_validation=True)
     connexion_app.add_api("mgmt_api.spec.yaml")
+    connexion_app.add_api("admin_api.spec.yaml", validate_responses=True, strict_validation=True)
```
CodeHeeler, line 27: In this file, you will also need to import similar to here in baselines:
https://github.com/RedHatInsights/system-baseline-backend/blob/master/system_baseline/app.py#L15

And you will need to register the blueprint as well I think:
https://github.com/RedHatInsights/system-baseline-backend/blob/master/system_baseline/app.py#L56

See this PR for further comparisons of how this was done for the internal endpoint:
https://github.com/RedHatInsights/system-baseline-backend/pull/179/files

drift/openapi/admin_api.spec.yaml
```diff
@@ -0,0 +1,36 @@
...
+  description: Admin API for Drift Service
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: "{{ path_prefix }}{{ app_name }}/admin"
+
+
+paths:
```
CodeHeeler, line 15: I feel like I recall the version being required, I don't know if there would be an exception for admin, but here is an example:
https://github.com/RedHatInsights/system-baseline-backend/blob/master/system_baseline/openapi/internal_api.spec.yaml#L14-L26

drift/openapi/admin_api.spec.yaml
```diff
@@ -0,0 +1,36 @@
...
+  /status:
+    get:
+      summary: admin data call
+      operationId: drift.views.admin.status
```
CodeHeeler, line 19: if so on my above comment about version, don't forget to add the version in here as well as `drift.admin_views.v1.status`

And I do think you need your own separate views like internal_views and mgmt_views, you would now have admin_views folder and v1.py within that.

drift/views/admin.py
```diff
@@ -0,0 +1,24 @@
+from flask import Blueprint, jsonify
```
CodeHeeler, line 1: Again, unless I am misunderstanding the new admin work, I think this file needs to be drift/admin_views/v1.py instead.

CodeHeeler (changes requested):
I've left a number of comments on things that need changing, but everything else looks great!  Thank you so much for working on this.

drift/views/admin.py
CodeHeeler: yeah, adding a comment to that effect is a good idea

drift/openapi/admin_api.spec.yaml
fellipeh: During the design process we decide to move the admin without any version, because admin will be used only by managers etc.. and will not have several changes

CodeHeeler: Ok, in that case, if it works locally and passes automation and schema checks, then that's fine.

drift/views/admin.py
CodeHeeler: given your decision to not have version, then this is probably drift/admin_views/admin.py for the naming convention then?

drift/app.py
fellipeh: fixed.

sthirugn (approved):
please wait for @CodeHeeler final review too

sthirugn:
nitpick: it will be great to add jira id in commit message

CodeHeeler (approved):
LGTM, thanks for considering my suggestions.

sthirugn (approved):

PR #279, opened 2021-08-06
[DRFT-197] Add reference based comparison for multivalues

CodeHeeler:

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -352,9 +352,9 @@ def _create_comparison(systems, info_name, reference_id, system_count, short_cir
                     reference_value = values["value"]
 
             for values in sorted_system_id_values_without_obfuscated:
-                values["state"] = COMPARISON_SAME
+                values["state"] = [COMPARISON_SAME]
                 if values["value"] != reference_value:
-                    values["state"] = COMPARISON_DIFFERENT
+                    values["state"] = [COMPARISON_DIFFERENT]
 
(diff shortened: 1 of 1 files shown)


CodeHeeler:
Summary:
Currently, the json we pass for comparison report is something like (totally rough and made up, bear with me, just trying to show what to look for):
{
...
"name": blah,
"state": "DIFFERENT", # <--this state is what we use for the state field and colored icons on the left of a comparison
"multivalues": blah, # present only when there are multi-value facts
"systems": [
{
"id": blah,
"value": blah
"state": "SAME" # <--until now, this state was only present on single value facts.
# this state is how the frontend decides which cell to shade red
},
{
"id": blah,
"value": blah
"state": "DIFFERENT"
}
]

So, the "state" inside of "systems" is now added on multivalue facts instead of only single value facts with this PR.  However for multivalue facts, this needs to be a list of states one for each item in the multivalue fact list, so as we did for another part of the code when we initially added multivalue facts, we will now make all of these "state" under "systems" a list so that the frontend doesn't have to check "isInstance" for list and complicate the code on their end.

PR #345, opened 2022-07-14
[DRFT-1137] Change Sort method

fellipeh:
That's my #1 (First) attempt to change sort method.. changing from `sorted()` to `.sort()`

after read tons of links, here what I found:

`sort()` is a method of the list class, and sorts the list in-place, returning None.
`sorted()` is a method built into the Python namespace, and sorts the list out-of-place, returning a sorted copy of the list, without affecting the original one.

in my tests, `sorted()` is more fast in small number of objects, but in large objects `sort` is more fast, something like this:

```
<100 elements, pretty much the same  ~0.5s
>=100 <1000 elements, `sort()` takes **half** of the time
>=1000 elements, `sorted()` takes nearly ten times as much computation time as `sort()` does.
```

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -99,7 +99,7 @@ def build_comparisons(
         grouped_comparisons = _group_comparisons(stripped_comparisons)
 
     with PT_BC_SORT_COMPARISONS.time():
-        sorted_comparisons = sorted(grouped_comparisons, key=lambda comparison: comparison["name"])
+        grouped_comparisons.sort(key=lambda comparison: comparison["name"])
 
     with PT_BC_CREATE_METADATA.time():
         # create metadata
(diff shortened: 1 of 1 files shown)


CodeHeeler (approved):
This looks good.  Looks like we don't reuse the original structures so there's no reason not to sort in place.  Thanks for this well researched and clear breakdown for your logic.

jpramos123 (approved):
LGTM!

Thank you for this investigation

PR #348, opened 2022-08-03
feat(DRFT-1137): Refactoring build_comparisons

fellipeh:
This refactor `build_comparisons`  method and all sub-method

## Secure Coding Checklist
- [ ] Input Validation
- [ ] Output Encoding
- [ ] Session Management
- [ ] Access Control
- [ ] Cryptographic Practices
- [ ] Data Protection
- [ ] Communication Security
- [ ] System Configuration
- [ ] Database Security
- [ ] File Management
- [ ] Memory Management

### DIFF
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -157,14 +157,6 @@ def _group_comparisons(comparisons):
     grouped_comparisons = []
     grouped_comparisons_with_dotnotation = {}
 
-    def _get_group_name(name):
-        n, _, _ = name.partition(".")
-        return n
-
-    def _get_value_name(name):
-        _, _, n = name.partition(".")
-        return n
-
     for comparison in comparisons:
         if "." in comparison["name"]:
             group_name, _, comparison_value_name = comparison["name"].partition(".")
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -170,9 +170,6 @@ def _group_comparisons(comparisons):
             if group["name"] == name:
                 return group
 
-    # with PT_GC_BUILD_GROUP_NAMES.time(): # move to proper place or remove competely
-    # with PT_GC_POPULATE_GROUPS.time(): # move to proper place or remove competely
-    # with PT_GC_SET_SUMMARY.time(): # move to proper place or remove competely
     for comparison in comparisons:
         if "." in comparison["name"]:
             group_name = _get_group_name(comparison["name"])
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -129,19 +129,18 @@ def build_comparisons(
     }
 
 
-def _set_summary_state(grouped_comparison):
-    states = {comparison["state"] for comparison in grouped_comparison["comparisons"]}
-    if COMPARISON_DIFFERENT in states:
-        grouped_comparison["state"] = COMPARISON_DIFFERENT
-    elif COMPARISON_INCOMPLETE_DATA in states:
-        if COMPARISON_SAME in states:
-            grouped_comparison["state"] = COMPARISON_DIFFERENT
+def summary_state(state_list):
+    if COMPARISON_DIFFERENT in state_list:
+        return COMPARISON_DIFFERENT
+    elif COMPARISON_INCOMPLETE_DATA in state_list:
+        if COMPARISON_SAME in state_list:
+            return COMPARISON_DIFFERENT
         else:
-            grouped_comparison["state"] = COMPARISON_INCOMPLETE_DATA
-    elif COMPARISON_SAME in states and len(states) == 1:
-        grouped_comparison["state"] = COMPARISON_SAME
-    else:  # use 'incomplete data' as the fallback state if something goes wrong
-        grouped_comparison["state"] = COMPARISON_INCOMPLETE_DATA
+            return COMPARISON_INCOMPLETE_DATA
+    elif COMPARISON_SAME in state_list and len(state_list) == 1:
+        return COMPARISON_SAME
+    # use 'incomplete data' as the fallback state if something goes wrong
+    return COMPARISON_INCOMPLETE_DATA
 
 
 def _group_comparisons(comparisons):
@@ -172,12 +171,6 @@ def _group_comparisons(comparisons):
             # modify existing group
             comparison["name"] = comparison_value_name
             group["comparisons"].append(comparison)
-
-            # trigger summary set action
-            # set summary state if grouped comparison contains groups
-            # as long as _set_summary_state doesn't rely on group["state"]
-            # we can trigger it more than once for every group
-            _set_summary_state(group)
         else:
             grouped_comparisons.append(comparison)
 
@@ -186,7 +179,9 @@ def _group_comparisons(comparisons):
         # sorting comparisons if exists
         group["comparisons"].sort(key=lambda comparison: comparison["name"])
 
-    # Need to add all dot notation comparisons into the output grouped_comparisons list
+        states = {comparison["state"] for comparison in group["comparisons"]}
+        group["state"] = summary_state(state_list=states)
+
     grouped_comparisons = grouped_comparisons + list(grouped_comparisons_with_dotnotation.values())
 
     return grouped_comparisons
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -172,9 +172,6 @@ def _group_comparisons(comparisons):
             # modify existing group
             comparison["name"] = comparison_value_name
             group["comparisons"].append(comparison)
-            group["comparisons"] = sorted(
-                group["comparisons"], key=lambda comparison: comparison["name"]
-            )
 
             # trigger summary set action
             # set summary state if grouped comparison contains groups
@@ -184,9 +181,13 @@ def _group_comparisons(comparisons):
         else:
             grouped_comparisons.append(comparison)
 
+    # sorting comparisons if exists
+    for group in grouped_comparisons_with_dotnotation.values():
+        # sorting comparisons if exists
+        group["comparisons"].sort(key=lambda comparison: comparison["name"])
+
     # Need to add all dot notation comparisons into the output grouped_comparisons list
-    for comparison in grouped_comparisons_with_dotnotation:
-        grouped_comparisons.append(grouped_comparisons_with_dotnotation.get(comparison))
+    grouped_comparisons = grouped_comparisons + list(grouped_comparisons_with_dotnotation.values())
 
     return grouped_comparisons
 
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -155,7 +155,7 @@ def _group_comparisons(comparisons):
 
     # default values for method run
     grouped_comparisons = []
-    processed_group_names = set()
+    grouped_comparisons_with_dotnotation = {}
 
     def _get_group_name(name):
         n, _, _ = name.partition(".")
@@ -165,29 +165,25 @@ def _group_comparisons(comparisons):
         _, _, n = name.partition(".")
         return n
 
-    def _find_group(name, grouped_comps):
-        for group in grouped_comps:
-            if group["name"] == name:
-                return group
-
     for comparison in comparisons:
         if "." in comparison["name"]:
-            group_name = _get_group_name(comparison["name"])
-            built_group = {"name": group_name, "comparisons": []}
+            group_name, _, comparison_value_name = comparison["name"].partition(".")
 
             # build out group names
-            if group_name not in processed_group_names:
-                processed_group_names.add(group_name)
-                grouped_comparisons.append(built_group)
+            if group_name not in grouped_comparisons_with_dotnotation:
+                built_group = {"name": group_name, "comparisons": []}
+                grouped_comparisons_with_dotnotation[group_name] = built_group
 
             # populate groups
-            group = _find_group(group_name, grouped_comparisons)
+            group = grouped_comparisons_with_dotnotation.get(group_name)
+
             # modify existing group
-            comparison["name"] = _get_value_name(comparison["name"])
+            comparison["name"] = comparison_value_name
             group["comparisons"].append(comparison)
             group["comparisons"] = sorted(
                 group["comparisons"], key=lambda comparison: comparison["name"]
             )
+
             # trigger summary set action
             # set summary state if grouped comparison contains groups
             # as long as _set_summary_state doesn't rely on group["state"]
@@ -195,6 +191,10 @@ def _group_comparisons(comparisons):
             _set_summary_state(group)
         else:
             grouped_comparisons.append(comparison)
+
+    # Need to add all dot notation comparisons into the output grouped_comparisons list
+    for comparison in grouped_comparisons_with_dotnotation:
+        grouped_comparisons.append(grouped_comparisons_with_dotnotation.get(comparison))
 
     return grouped_comparisons
 
--- a/drift/info_parser.py
+++ b/drift/info_parser.py
@@ -129,6 +129,21 @@ def build_comparisons(
     }
 
 
+def _set_summary_state(grouped_comparison):
+    states = {comparison["state"] for comparison in grouped_comparison["comparisons"]}
+    if COMPARISON_DIFFERENT in states:
+        grouped_comparison["state"] = COMPARISON_DIFFERENT
+    elif COMPARISON_INCOMPLETE_DATA in states:
+        if COMPARISON_SAME in states:
+            grouped_comparison["state"] = COMPARISON_DIFFERENT
+        else:
+            grouped_comparison["state"] = COMPARISON_INCOMPLETE_DATA
+    elif COMPARISON_SAME in states and len(states) == 1:
+        grouped_comparison["state"] = COMPARISON_SAME
+    else:  # use 'incomplete data' as the fallback state if something goes wrong
+        grouped_comparison["state"] = COMPARISON_INCOMPLETE_DATA
+
+
 def _group_comparisons(comparisons):
     """
     given a list of comparisons, group similar names. For example:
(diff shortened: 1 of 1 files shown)


drift/info_parser.py
```diff
@@ -150,46 +151,37 @@ def _get_value_name(name):
...
-        for group in grouped_comps:
-            if group["name"] == name:
-                return group
-
     def _set_summary_state(grouped_comparison):
-        states = {comparison["state"] for comparison in grouped_comparison["comparisons"]}
+        states = {comparison["state"] for comparison in grouped_comparison}
         if COMPARISON_DIFFERENT in states:
-            grouped_comparison["state"] = COMPARISON_DIFFERENT
+            grouped_comparison[0]["state"] = COMPARISON_DIFFERENT
```
vkrizan, line 157: What's the magic `0`? :-)

vkrizan (commented):
Check comments.

Would it be possible to first write a unit tests (with couple of cases) and then change the algorithm?

drift/info_parser.py
```diff
@@ -140,6 +140,7 @@ def _group_comparisons(comparisons):
 
     # default values for method run
     grouped_comparisons = []
+    temp_grouped_comparisons_dict = {}
```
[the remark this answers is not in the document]
fellipeh: We can't transform a dict into a list, just "transforming" it.. in our case we need to create a new list and iterate into that dict passing the new values, since dict we don't have the same output format we had in the list.

drift/info_parser.py
fellipeh: Because `get` on the dict, returns a list of values. Here is one sample:
```
[{'name': 'bash', 'state': 'DIFFERENT', 'systems': [{'id': 'ff35596c-f98e-11e9-aea9-98fa9b07d419', 'value': ''}, {'id': '89df6310-f98e-11e9-8a65-98fa9b07d419', 'value': ''}, {'id': 'fc1e497a-28ae-11e9-afd9-c85b761454fa', 'value': '4.4.23-6.fc29.x86_64'}, {'id': 'bbbbbbbb-28ae-11e9-afd9-c85b761454fa', 'value': '4.4.23-6.fc29.x86_64'}, {'id': 'bbbbbbbb-28ae-11e9-afd9-c85b761454fa', 'value': '4.4.23-6.fc29.x86_64'}], 'drifted_from_baseline': False}]
```

it's a list of only 1 result, with dict inside it, that's the reason to get [0]

drift/info_parser.py
```diff
@@ -140,6 +140,7 @@ def _group_comparisons(comparisons):
...
     processed_group_names = set()
```
vkrizan, line 144: Is `processed_group_names` sill necessary?

drift/info_parser.py
```diff
@@ -150,46 +151,37 @@ def _get_value_name(name):
         _, _, n = name.partition(".")
         return n
 
-    def _find_group(name, grouped_comps):
-        for group in grouped_comps:
-            if group["name"] == name:
-                return group
-
     def _set_summary_state(grouped_comparison):
```
vkrizan, line 154: Shadowed variable `grouped_comparison`.

drift/info_parser.py
vkrizan: this does not look good @fellipeh. What happens if there is more than one comparison group?

vkrizan:
@fellipeh Here are my recommendations. 

First lets start over, from the original implementation.
Then try to improve the following places/steps one by one, each in a separate commit (incremental changes):
1. Remove `_find_group` and replace it with dict
   * remove unnecessary sets and variables along
2. Sort comparisons once after they are grouped
3. Move `_set_summary_state` out to be an normal function (not an inner one)
   * in: states; out: state
4. Try to remove the else block completely and use dict exclusively (even-though the grouped ones are not that common)

For each of the steps ensure that the algorithm works as expected.
Utilize existing objects, but do not hesitate to copy them as necessary. Avoid changing input objects. Avoid adding and appending to random objects, as the `[0]` noted in https://github.com/RedHatInsights/drift-backend/pull/348#discussion_r937722397.

Note that even having an exclusive dict implemented, we can still profile it and revert it as necessary. Thanks to separate commits.

Hopefully this gives you a better angel.

vkrizan:
Another one to the list:
5. Partition the name to parent and group once.

drift/info_parser.py
```diff
@@ -99,7 +99,7 @@ def build_comparisons(
         grouped_comparisons = _group_comparisons(stripped_comparisons)
 
     with PT_BC_SORT_COMPARISONS.time():
-        grouped_comparisons.sort(key=lambda comparison: comparison["name"])
+        grouped_comparisons.sort(key=lambda comparison: "name")
```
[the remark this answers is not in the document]
fellipeh: rolled back

drift/info_parser.py
fellipeh: fixed

drift/info_parser.py
fellipeh: fixed

drift/info_parser.py
vkrizan: @fellipeh this one is still relevant

drift/info_parser.py
```diff
@@ -129,6 +129,21 @@ def build_comparisons(
     }
 
 
+def _set_summary_state(grouped_comparison):
```
vkrizan, line 132: Can you please also refactor it to take the list of states and return a computed state? It would read much better and would not affect the input data structure.

drift/info_parser.py
```diff
@@ -196,6 +194,10 @@ def _find_group(name, grouped_comps):
         else:
             grouped_comparisons.append(comparison)
 
+    # Need to add all dot notation comparisons into the output grouped_comparisons list
+    for comparison in grouped_comparisons_with_dotnotation:
+        grouped_comparisons.append(grouped_comparisons_with_dotnotation.get(comparison))
```
vkrizan, line 199: Is there a way to append it without a `for` loop? This adds another cycles.

vkrizan (commented):
Hey @fellipeh, the direction is good. 
There are still some steps to be done:
* sorting once ([2](https://github.com/RedHatInsights/drift-backend/pull/348#issuecomment-1205528751)) 
* removing the else block ([4](https://github.com/RedHatInsights/drift-backend/pull/348#issuecomment-1205528751))
* partitioning the name once ([5](https://github.com/RedHatInsights/drift-backend/pull/348#issuecomment-1206159713))

But most imporantly there should be unittest written **before** any new changes to verify the algorithm!

drift/info_parser.py
fellipeh: Fixed here: a4285a58dac2122380078195dd6d94d8173f9366

fellipeh:
Hey @vkrizan 

>     * sorting once ([2](https://github.com/RedHatInsights/drift-backend/pull/348#issuecomment-1205528751))

Fixed on this commit: b6a073ca578f49a6579edac7da025001b051f51c

>     * removing the else block ([4](https://github.com/RedHatInsights/drift-backend/pull/348#issuecomment-1205528751))

Working on it..

>     * partitioning the name once ([5](https://github.com/RedHatInsights/drift-backend/pull/348#issuecomment-1206159713))

Done, on this commit: d558733a9c36a4c83349b5996aaad2c27bf8c00c

drift/info_parser.py
```diff
@@ -187,6 +184,12 @@ def _group_comparisons(comparisons):
         else:
             grouped_comparisons.append(comparison)
 
+    # sorting comparisons if exists
+    for comparison_name in grouped_comparisons_with_dotnotation:
+        group = grouped_comparisons_with_dotnotation.get(comparison_name)
+        group["comparisons"] = sorted(
```
vkrizan, line 190: Can it be done without a for loop (or at least without the get by key) and sorted inplace?

drift/info_parser.py
```diff
@@ -190,9 +190,11 @@ def _group_comparisons(comparisons):
         group["comparisons"] = sorted(
             group["comparisons"], key=lambda comparison: comparison_name
         )
+
+    grouped_comparisons = grouped_comparisons + list(grouped_comparisons_with_dotnotation.values())
     # Need to add all dot notation comparisons into the output grouped_comparisons list
-    for comparison in grouped_comparisons_with_dotnotation:
-        grouped_comparisons.append(grouped_comparisons_with_dotnotation.get(comparison))
+    # for comparison in grouped_comparisons_with_dotnotation:
+    #     grouped_comparisons.append(grouped_comparisons_with_dotnotation.get(comparison))
```
vkrizan, line 197: Please remove these. I suggest to fixup this change to the corresponding commit: "Using dict instead of list".

drift/info_parser.py
```diff
@@ -140,56 +155,27 @@ def _group_comparisons(comparisons):
...
-            group_name = _get_group_name(comparison["name"])
+            group_name, _, comparison_value_name = comparison["name"].partition(".")
+
             built_group = {"name": group_name, "comparisons": []}
 
             # build out group names
             if group_name not in processed_group_names:
                 processed_group_names.add(group_name)
-                grouped_comparisons.append(built_group)
+                grouped_comparisons_with_dotnotation.update({group_name: built_group})
```
vkrizan, line 170: Can this be done without calling update?

drift/info_parser.py
```diff
@@ -140,56 +155,27 @@ def _group_comparisons(comparisons):
...
-
-    # with PT_GC_BUILD_GROUP_NAMES.time(): # move to proper place or remove competely
-    # with PT_GC_POPULATE_GROUPS.time(): # move to proper place or remove competely
-    # with PT_GC_SET_SUMMARY.time(): # move to proper place or remove competely
     for comparison in comparisons:
         if "." in comparison["name"]:
-            group_name = _get_group_name(comparison["name"])
+            group_name, _, comparison_value_name = comparison["name"].partition(".")
+
             built_group = {"name": group_name, "comparisons": []}
```
vkrizan, line 165: Does this need to be created each time for a grouped comparisons?

drift/info_parser.py
fellipeh: I believe not, since we're sorting a child item, we need to pick up the object and sort it.

drift/info_parser.py
fellipeh: I'm afraid not, since we're adding a new `dict`, the `update` is the method to include it.

drift/info_parser.py
fellipeh: Yes, since the group name and value can be changed. Our input is not sorted, so, we need to assume different values from the `comparison` for each item.

drift/info_parser.py
fellipeh: That's what this code does. :)
```
states = {comparison["state"] for comparison in grouped_comparison["comparisons"]}
```

Here return a list of state in that group, and return computed state after.

drift/info_parser.py
fellipeh: fixed

fellipeh:
Hey @vkrizan if you're satisfied with all the answers, we can proceed to merge into stage to start to see if this make some difference.

The only thing I did do it is to remove the else, as we discussed, we can do it in another PR.

drift/info_parser.py
vkrizan: How about direct assignment by key?
```
grouped_comparisons_with_dotnotation[group_name] = ...
```

drift/info_parser.py
vkrizan: Please check where the `built_group` is referenced.

```
            built_group = {"name": group_name, "comparisons": []}

            # build out group names
            if group_name not in processed_group_names:
                processed_group_names.add(group_name)
                grouped_comparisons_with_dotnotation.update({group_name: built_group})
```
Can it be move do the `if` block?

drift/info_parser.py
vkrizan: This function does a side effect on the input. My suggestion is to do a return value and no side effect.

State gathering should move out
```
states = {comparison["state"] for comparison in grouped_comparison["comparisons"]}
```

and the function should look like this:
```
def summary_state(states):
    if COMPARISON_DIFFERENT in states:
        return COMPARISON_DIFFERENT
    elif COMPARISON_INCOMPLETE_DATA in states:
    ...
``` 

then outside of the function
```
group_states = {comparison["state"] for comparison in group["comparisons"]}
group["state"] = summary_state(group_states)
```

drift/info_parser.py
vkrizan: Please fix it up with commit "Using dict instead of list". The issue was introduced in this PR so it does not need to be carried over to main branch.

drift/info_parser.py
vkrizan: This for loop
```
for comparison_name in grouped_comparisons_with_dotnotation:
```
gives keys.

Can it be changed to get `.values()` only and sort inplace? The similar way as you recently did? List has a sort method. `sorted` is a generic builtin and creates a new list. We should reduce the number of objects being created.

To make it without a for-loop, there are comperhensions which can be quicker with the cost of readability.

vkrizan:
@fellipeh Please go through it again. There is one more comment that was not been addressed: https://github.com/RedHatInsights/drift-backend/pull/348/commits/f8954dedb3bca3200fbdc1a9ec144f6419be72a1#r937920319

Plus I've noticed that the `_set_summary_state` is called too many times. Can we move to the loop that does the sorting?

Please address the comments on the code not just by replying to it. There is an answer for each of my questions.

drift/info_parser.py
fellipeh: Make sense, fixed on this commit: c46c812f9e26e5d42f939b7e5b8dd7b256b365d7

drift/info_parser.py
fellipeh: I tested here, and to be honest doesn't take much affect into the time.

And, using direct assignment have a side affect also (look at the screenshots).. **only in that case**, we don't need the `dict` with a key.. we need a dict without a key to build the comparisons. We can move with the direct assignment, but to do that, we need to change all the logic in the code, I don't know if the result compensates for all this change. 

Using update:

Screen Shot 2022-08-10 at 10 38 55

Using direct assignment:
Screen Shot 2022-08-10 at 10 35 07

drift/info_parser.py
fellipeh: Fixed here: c9557a969b10fd3073dda1c651bb7d573867aef5

drift/info_parser.py
fellipeh: Yes, we need to check if the group was already created, we can't move the `get` on dict (https://github.com/RedHatInsights/drift-backend/pull/348/files#diff-d94d1e0636fff7ab5a4d4f6f24ff091770a8a2ec2c170453ec644766a4f16857R171) to above it to test, since we need to get the item after the creation.. so, moving it, we need to have a second get into the dict, which is more thing to do... so.. using only on `set` to check it, is more efficient, I believe.

drift/info_parser.py
```diff
@@ -175,20 +174,22 @@ def _group_comparisons(comparisons):
...
+        # sorting comparisons if exists
         group["comparisons"] = sorted(group["comparisons"], key=lambda comparison: comparison_name)
 
+        # trigger summary set action
+        # set summary state if grouped comparison contains groups
+        # as long as summary_state doesn't rely on group["state"]
+        # we can trigger it more than once for every group
+        states = {comparison["state"] for comparison in group["comparisons"]}
+        group["state"] = summary_state(state_list=states)
+        summary_state(group)
```
vkrizan, line 190: This line should be removed.

drift/info_parser.py
vkrizan: Set and dict are formally the same with regards to key existence checks. The same check can be use with the dict that is being populated, cannot it?

drift/info_parser.py
vkrizan: Sorry I do not understand the reasoning. Direct assignment and update would have the same result. The only difference is that for the update a new object has to be created that is iterable, then iterated on and then added to dictionary. I acknowledge that the save on speed would not be much, however with direct assignment it reads better and uses one less objects that needs to be GCd.

```
>>> d = {}
>>> built_group = {"name": "g1", "comparisons": []}
>>> d.update({"g1": built_group})
>>> built_group = {"name": "g2", "comparisons": []}
>>> d["g2"] = built_group
>>> d
{'g1': {'name': 'g1', 'comparisons': []}, 'g2': {'name': 'g2', 'comparisons': []}}
```

drift/info_parser.py
```diff
@@ -187,6 +184,12 @@ def _group_comparisons(comparisons):
         else:
             grouped_comparisons.append(comparison)
 
+    # sorting comparisons if exists
+    for comparison_name in grouped_comparisons_with_dotnotation:
+        group = grouped_comparisons_with_dotnotation.get(comparison_name)
+        group["comparisons"] = sorted(
+            group["comparisons"], key=lambda comparison: comparison_name
```
vkrizan: CI fails here. Check where the `comparison_name` comes from within the lambda and you'll find a solution.

drift/info_parser.py
fellipeh: Fixed here: 40f4d1843e8c43cb2b7f2b2244d2c8e3d3ffc386

drift/info_parser.py
```diff
@@ -140,64 +154,42 @@ def _group_comparisons(comparisons):
...
         if "." in comparison["name"]:
-            group_name = _get_group_name(comparison["name"])
-            built_group = {"name": group_name, "comparisons": []}
+            group_name, _, comparison_value_name = comparison["name"].partition(".")
 
             # build out group names
-            if group_name not in processed_group_names:
-                processed_group_names.add(group_name)
-                grouped_comparisons.append(built_group)
+            if group_name not in grouped_comparisons_with_dotnotation.keys():
```
vkrizan, line 164:
```suggestion
            if group_name not in grouped_comparisons_with_dotnotation:
```

drift/info_parser.py
```diff
@@ -140,64 +154,42 @@ def _group_comparisons(comparisons):
...
+                built_group = {"name": group_name, "comparisons": []}
+                grouped_comparisons_with_dotnotation.update({group_name: built_group})
```
vkrizan, line 166:
```suggestion
                grouped_comparisons_with_dotnotation[group_name] = built_group
```

drift/info_parser.py
```diff
@@ -140,64 +154,42 @@ def _group_comparisons(comparisons):
...
-            _set_summary_state(group)
         else:
             grouped_comparisons.append(comparison)
 
+    for comparison_name in grouped_comparisons_with_dotnotation:
+        group = grouped_comparisons_with_dotnotation[comparison_name]
+        # sorting comparisons if exists
+        group["comparisons"] = sorted(
+            group["comparisons"], key=lambda comparison: comparison["name"]
+        )
```
vkrizan, line 182:
```suggestion
        group["comparisons"].sort(key=lambda comparison: comparison["name"])
```

drift/info_parser.py
```diff
@@ -140,64 +154,42 @@ def _group_comparisons(comparisons):
...
-            # trigger summary set action
-            # set summary state if grouped comparison contains groups
-            # as long as _set_summary_state doesn't rely on group["state"]
-            # we can trigger it more than once for every group
-            _set_summary_state(group)
         else:
             grouped_comparisons.append(comparison)
 
+    for comparison_name in grouped_comparisons_with_dotnotation:
+        group = grouped_comparisons_with_dotnotation[comparison_name]
```
vkrizan, line 178:
```suggestion
    for group in grouped_comparisons_with_dotnotation.values():
```

vkrizan (approved):

sthirugn:
yay :+1:

PR #356, opened 2022-10-12
Add apiPath to webservice for frontend

adamrdrew:
This patch adds apiPath to the webservice for the backend API in the Clowdapp. This ensures the new containerized frontend app knows how to reach the API.


petrblaho:
Thank you, @adamrdrew .

Can you please link a ticket this PR is going to solve?

Also, reformatting of the complete file in the same commit as the feature makes it almost unreviewable. I know that our `clowdapp.yaml` is not nicely formatted so I created https://github.com/RedHatInsights/drift-backend/pull/357 . When this is merged you can rebase this PR so the change is clearly visible.

Also I guess that the same change will be needed in https://github.com/RedHatInsights/system-baseline-backend and in https://github.com/RedHatInsights/historical-system-profiles-backend as these are two other backend apps that are part of Drift and frontend should be pulling data from them as well.

adamrdrew:
Hello! Thanks for the response. This will help close this issue:
https://issues.redhat.com/browse/RHCLOUD-20273

I know the file re-format is a pain; For some context:  I use a tool to do these migrations and it programmatically re-writes the files. (It deserializes the clowdapp YAML, makes the change to the object, and then reserializes it.) The process for doing the frontend migrations requires more than 10 changes to files from 4 repos per app, so I need to do it programmatically or it would just take too long and be too error prone if I did it all manually :) The only real functional change in this PR is the addition of `apiPath: drift` to the public web service so the frontend can find it.

Thanks for the heads up on those other repos - every other app I've touched has only had one backend so I've never seen this before. I'll need to look at the frontend code and see if they are fetching from these other sources and if so, you're right I'll need to add apiPaths. Thanks!

sthirugn:
@adamrdrew can you please rebase? our clowdapp.yaml format is now fixed and merged.

petrblaho:
@adamrdrew I rebased your change, I hope you don't mind.

I will merge when the CI passes.

petrblaho (approved):

PR #639, opened 2024-04-03
feat: updates dockerfile

petrblaho:
RHINENG-1755

Now we use multistage build,
python39 stream module,
and we unified dockerfiles content accross drift services (expect deps).

This way the resulting image is cca 50% smaller.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,89 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
+ARG deps="git-core python39-pip tzdata"
+ARG buildDeps="python39-devel gcc"
+ARG poetryVersion="1.6.0"
+
+ARG TEST_IMAGE=false
+
+#######################
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal AS base
+
+ARG deps
+ARG poetryVersion
+
+ENV LC_ALL=C.utf8
+ENV LANG=C.utf8
 
 RUN microdnf update -y && \
-    microdnf install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs -y \
-    git-core python39 python39-pip tzdata libpq-devel && \
-    rpm -qa | sort > packages-before-devel-install.txt && \
-    microdnf install --setopt=tsflags=nodocs -y python39-devel gcc && \
-    rpm -qa | sort > packages-after-devel-install.txt
+    microdnf module enable python39 && \
+    microdnf install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs -y $deps && \
+    microdnf clean all
+RUN pip3 install --force-reinstall poetry~="${poetryVersion}"
 
-RUN adduser --gid 0 -d /opt/app-root --no-create-home insights
+#######################
+
+FROM base AS build
+
+ARG buildDeps
+ARG poetryVersion
 
 ENV LC_ALL=C.utf8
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,39 +1,76 @@
...
+
+RUN microdnf update -y && \
+    microdnf module enable python39 && \
+    microdnf install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs -y $deps
+RUN pip3 install --force-reinstall poetry~=1.6.0
 
-RUN chown -R insights:0 /opt/app-root  && \
-    chgrp -R 0 /opt/app-root && \
-    chmod -R g=u /opt/app-root
+RUN adduser --gid 0 -d /opt/app-root --no-create-home insights
```
vkrizan, line 63: Hmm, what is going on here? :thinking: Why it uses root's group?

Dockerfile
```diff
@@ -1,39 +1,76 @@
...
+    POETRY_DATA_DIR=/opt/app-root/.pypoetry/data \
+    POETRY_CACHE_DIR=/opt/app-root/.pypoetry/cache
+
+ENV VIRTUAL_ENV_DIR=${APP_ROOT}/src/.venv
+ENV UNLEASH_CACHE_DIR=/tmp/unleash_cache
+
+RUN microdnf update -y && \
+    microdnf module enable python39 && \
+    microdnf install --setopt=install_weak_deps=0 --setopt=tsflags=nodocs -y $deps
+RUN pip3 install --force-reinstall poetry~=1.6.0
```
vkrizan, line 61: If we're heavily using arguments and env vars here, can we move the poetry version up to the top of the file?

Dockerfile
```diff
@@ -1,39 +1,76 @@
...
-    chgrp -R 0 /opt/app-root && \
-    chmod -R g=u /opt/app-root
+RUN adduser --gid 0 -d /opt/app-root --no-create-home insights
 
 USER insights
 
-RUN poetry install --without dev --sync
+WORKDIR ${APP_ROOT}/src
+
+COPY --chown=insights:0 . ${APP_ROOT}/src
```
vkrizan, line 69: Again, what's up with the root group usage?

Dockerfile
```diff
@@ -1,39 +1,76 @@
...
+
+COPY --from=build --chown=insights:0 $VIRTUAL_ENV_DIR $VIRTUAL_ENV_DIR
+
+# allows unit tests to run successfully within the container if image is built in "test" environment
+RUN if [ "$TEST_IMAGE" = "true" ]; then chgrp -R 0 $APP_ROOT && chmod -R g=u $APP_ROOT; fi
```
vkrizan, line 74: Can the tests be updated in the fashion of having them successfully run withought chgrp/chmod mangling in this file?

vkrizan: root user should be avoided for applications. `USER 0` is there to install packages, after that there is a flip to a normal user (`USER 1001`).

vkrizan: App folder can be set to be under root group to have it readable/writable by a random user (UID) that OpenShift/k8s sets when running containers. (ref: https://developers.redhat.com/articles/2021/11/11/best-practices-building-images-pass-red-hat-container-certification)

Dockerfile
petrblaho: OpenShift ...

vkrizan (approved):
Noice!

